PR #479, opened 2015-01-06
Trigger groups!!

kadst43:


kadst43:
@AmbientOSX

Implements trigger groups.  Also cleans up some of the places we were using the blackboard to prevent repetition, replacing them with the doOnce clause.

Tested, and this does fix the issue where both success and failure fire simultaneously.

source/scenario/globalScenarioTriggers.vwf.yaml
```diff
@@ -127,14 +118,13 @@ properties:
         - musicFailure
 
     failOnCollision:
+      group: successOrFailure
+      priority: 2
       triggerCondition:
       - and:
```
AmbientOSX: No need for an AND here

source/scenario/globalScenarioTriggers.vwf.yaml
```diff
@@ -178,14 +168,13 @@ properties:
           - ALVO32_Rover
 
     failOnBatteryDepleted:
+      group: successOrFailure
+      priority: -2
       triggerCondition:
       - and:
```
AmbientOSX: No AND needed

AmbientOSX:
Rest looks :+1:

PR #480, opened 2015-01-12
Text to speech work

kadst43:
Switches over to using text-to-speech.

@AmbientOSX @BrettASwift @nmarshak1337


source/marsGameSound.vwf.yaml
```diff
@@ -20,6 +20,12 @@ properties:
   soundSet:
     - soundName: "musicStandardGameplay"
       soundURL: "assets/sounds/music/musicStandardGameplayLoopFullMix.ogg"
+      useTextToSpeech: true
+      ttsAmplitude: 100
+      ttsVariant: "m1"
+      ttsWordGap: 0
+      ttsSpeed: 175
+      ttsPitch: 50
```
kadst43: Let's move these into voices.

kadst43: Also, merge amplitude into the existing volume control (but maybe let the voice set the base amplitude - or something).

kadst43: Also, add warnings if you specify text-to-speech but no line is given (go off the actual text-to-speech line, not the subtitle), or if a sound file is specified on a text-to-speech-line.

Also, separate out the string that we generate the speech from, instead of using the subtitle.

kadst43:
Also, the changes with respect to when the buffer gets created.

index.vwf.html
```diff
@@ -173,7 +174,7 @@
       <value name="HIGH">
         <block type="math_number">
           <field name="NUM">100</field>
-        </block>
+        </block>ex.vwf.html:<scri
```
AmbientOSX: Spooky stuff going on here..

nmarshak1337:
Closing because meSpeak uses GPL. We will revisit text-to-speech later (and reuse some of this code, potentially).

PR #481, opened 2015-01-13
Menu in Model + Level (area) Manager

BrettASwift:
@AmbientOSX @kadst43 

While I was trying to find the "best way" to get the main menu into the model for replication, I kept coming up with essentially the same functionality. This branch was made with the thought that a level had scenarios and a physical scene, but I never got to the point where a level controlled the scenarios (which is the biggest difference in what I was doing and our Mission/Mission-Task idea). So, this as it stands, is just allowing us to swap out the physical scene. I am sure it will need modification, but this does give us that basic ability to switch between scenes/areas. I'll probably want to go through it and rename some things for clarity if it seems like what we want.

Please, look it over and let me know if anything stands out as not being useful. I already tested the trigger-improvements branch on top of this branch to make sure everything will work, and it does. I did not test replication.


AmbientOSX:
I rebased off of trigger-improvements and ran this. Still seeing strange behavior with the menu screen. If you load one client and then load another with that same URL and start the game on the second client you'll see the main menu text but not the overlay. I need to look through the changes and see what might be causing that...

BrettASwift:
Gotcha. I'll do some testing. It's probably just that we don't have anything in the model telling the view what UI should be displayed. I'll see what I can do about that.

BrettASwift:
Closing this for now. After a closer analysis, some of this doesn't really fit in with the Mission system idea and might cause us some headaches with area loading, so I'm going to do something more basic and focused on just the main menu and the replication of the menu and UI.

PR #482, opened 2015-01-15
Application State

BrettASwift:
@AmbientOSX @kadst43 

This is still a work in progress. I have a couple of bugs to look into and I'm making sure the replication works in regard to the menu before calling it "done."

First, I added the main menu to the model side and modified the menu view code to match. I also added the `applicationState` property to the scene which is handled in `scene.js` to set up the model side and in `index.js` (under `vwf_view.satProperty`) to handle the view side. The `loading` state is for a potential loading screen.

I wanted to get this up so it can be reviewed and I can make changes as I'm working on it tomorrow and this weekend, if I'm still on it.


kadst43:
Seems reasonable to me.

BrettASwift:
Rebased off of trigger improvements + prop cache. Resubmitting

PR #484, opened 2015-01-16
DONT MERGE: Blockly proc and vars

AmbientOSX:
@kadst43 A VWF dev/mars-game-staging rebase is forthcoming. Coordinating to make sure no GENI servers are affected.


source/grid.js
```diff
@@ -18,22 +18,24 @@
...
+            tiles[ i ][ j ] = new GridTile();
         }
     }
     this.future( 0 ).setBoundaryValues();
 }
 
 this.setBoundaryValues = function() {
-    if ( this.boundaryValues ) {
+    var boundaryValues = this.boundaryValues;
+    var tiles = this.tiles;
```
kadst43: Why copy these into local variables?  Is it just so you can avoid saying "this." in the rest of the code?  If so, I would suggest *not* doing it, because it is considerably less clear (the "this." signals to me that I'm looking at a member of the object, rather than a local variable).

Same comment throughout, obviously.

source/scenario/introScreenScenario.vwf.yaml
```diff
@@ -17,7 +17,7 @@ extends: source/scenario/scenario.vwf
 properties:
   scenarioName: introScreenScenario
 
-  nextScenarioPath: "scenario1a"
+  nextScenarioPath: "scenario1h"
```
kadst43: Presumably you don't want to check this in?

source/grid.js
AmbientOSX: Presumably this ties into some of David's comments about replication. I didn't make these changes.

source/scenario/introScreenScenario.vwf.yaml
AmbientOSX: This is just a branch to demo the blockly functionality on Tuesday - definitely dont want to check this in haha :)

kadst43: Ahhh... let's not do this.  I can jump ahead using the option in settings if I need to.

source/grid.js
kadst43: @davideaster Is this just a convenience, or does it need to be this way?

davideaster: Since `setBoundaryValues` is a property, each reference is a trip through the kernel. Keeping a local reference lets the iteration work with just one property access.

The mapping that makes node references work happens on every property reference. Since node mapping now walks through the entire property, it's important to avoid redundant access to properties that contain large objects.

You could avoid the explicit local reference for `setBoundaryValues` (but not `tiles`) by iterating like this:

    this.setBoundaryValues = function() {
        var tiles = this.tiles;
        ( this.boundaryValues || [] ).forEach( function( boundaryValuesX, i ) {
            boundaryValuesX && boundaryValuesX.forEach( function( boundaryValuesXY, j ) {
                tiles[ i ][ j ].energyRequired = boundaryValuesXY;
            }, this );
        }, this );

kadst43: Got it.  Makes sense.  Thanks!

kadst43:
So... these changes seem like they may be worth merging back in... except for the xml changes that actually make the blocks available to the player, and the change that starts the game on scenario 1h.

PR #485, opened 2015-01-19
Application state

BrettASwift:
@kadst43 @AmbientOSX 

I rebased this branch on trigger-improvements+property-caching, since we will need that branch to work on vwf/development.


source/mainMenu.js
```diff
@@ -0,0 +1,43 @@
...
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS, 
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+this.setup = function() {
+    this.scene.gameCam.setCameraTarget( this.target );
+    this.scene.soundManager.playSound( "musicMenuIntro" );
+    this.scene.soundManager.playSound( "musicMenuLoop" );
```
kadst43: Isn't this handled by the main menu scenario?

source/scene.js
```diff
@@ -32,7 +32,64 @@ this.initialize = function() {
...
+            this.environment.visible = false;
+            this.player.visible = false;
+            this.airDust.visible = false;
+            this.smoke1.visible = false;
+            this.smoke2.visible = false;
+            this.smoke3.visible = false;
+            this.backdrop.visible = false;
+            this.sunLight.visible = false;
+            this.envLight.visible = false;
+            this.pickups.visible = false;
```
kadst43: Since you appear to be setting all of these to true/false, does it make sense to extract this out to a function somewhere or something?  Might be handy if we ever wanted to do the same thing somewhere else (like when loading a new area).  Just a thought...

source/mainMenu.js
BrettASwift: There is no main menu scenario with these changes. It doesn't make sense anymore, particularly with the upcoming Mission system.

source/scene.js
BrettASwift: As the `TODO` suggests, I will likely place these under a parent node and just set the visibility of that node instead. Since these are all objects that would exist in a single area, they will likely end up in an `areaNode` or something.

kadst43:
:+1: 

It all looks reasonable to me.

BrettASwift:
This branch works with https://github.com/virtual-world-framework/vwf/pull/430. I'm using these two to bring mars-game up to vwf/development.

AmbientOSX:
Hmm ok. I'll have to wrap my head around utilizing that for future replication changes. At this point we should probably get this branch to a point where it works for 1 client without errors and merge it in I suppose...

BrettASwift:
@AmbientOSX I'm making some fixes right now. I'm going to make some comments to point out what's wrong, but I will be adding the fixes as well, so don't worry about addressing the comments.

AmbientOSX:
Isnt that what you told me to do? Call the method from the trigger and then handle either setting the property or modifying the view in scene.js?

BrettASwift:
Quick aside: I hate how GitHub line comments are all added to the main comments... Hard to respond to a specific comment...

Anyway, no, that's not what I was saying. Sorry for the confusion. I was saying to set the property in the model and then handle it on the view side in `vwf_view.satProperty`, which you did. However, the view function call in the `default` case of those switch statements is not the way to go. If we don't have a specific case, then we need to log a warning.

BrettASwift:
Closing this. Trying to get everything up to date and pushed to a new branch `mars-game-development`.

PR #486, opened 2015-01-26
Staging to Master

BrettASwift:
@kadst43 @AmbientOSX 

It looks like the camera system is the only difference here. It's probably not necessary to merge this. I'll leave that decision to you, Kevin.


kadst43:
Assuming that we're comfortable that these changes are stable and working, let's merge them up.  That much less to go through next time.

kadst43:
Test it, but :+1: 

I don't think we need to PR this kind of merge, since it already got reviewed on the way into staging.

PR #487, opened 2015-01-31
Dummy scenario with three rovers

nmarshak1337:
We can now play with three rovers in a dummy scenario (choose scenario_dummy, which comes before mainMenuScenario).


index.vwf.config.yaml
```diff
@@ -23,6 +23,6 @@ view:
   vwf/view/threejs: "#vwf-root"
   vwf/view/blockly: { "blocklyPath": './source/blockly/', "toolbox": 'toolbox' }
   vwf/view/sound:
-  # vwf/view/editor:
+  vwf/view/editor:
```
BrettASwift: Re-disable this. We never want the editor enabled on a public version of the game.

nmarshak1337:
Now we have multiple rovers in a dummy scenario, with Spencer's additions (procedures and variables) incorporated, along with all of our original blocks. 

Known Issues:
- This breaks tracing, but Kevin Dill gave me the goahead. 
- In the dummy scenario, switch to Rover 2, then back to Rover 1, for the "drive" block to NOT be grayed out.

kadst43:
Just to confirm... all blocks should be available, not just the drive block, right?

nmarshak1337:
@kadst43 Per the discussion we just had, I need to bolt on the graph tool, plus add a few more things to the toolbox (everything in the graph tool plus y=, also make sure that everything that we use in the game actually goes into the toolbox.

kadst43:
And a kitchen sink.  Don't forget the kitchen sink. ;)

You might close this PR, and then re-open when it's ready...

PR #489, opened 2015-02-04
Fixes blockly tabs - breaks blockly visibility

AmbientOSX:
@BrettASwift Any ideas?

### DIFF
--- a/source/scene.js
+++ b/source/scene.js
@@ -394,10 +394,11 @@ this.playVideo = function( src ) {
     playVideo( id );
 }
 
-this.enableBlocklyTab = function ( objectID ) {
-    var tabs = this.blocklyTabs;
-    tabs.push( objectID );
-    this.blocklyTabs = tabs;
+this.enableBlocklyTab = function ( nodeID ) {
+    var node = this.findByID( this, nodeID );
+    if ( node ) {
+        node.UIEnabled = true;
+    }
 }
 
 //@ sourceURL=source/scene.js
--- a/source/scene.js
+++ b/source/scene.js
@@ -394,4 +394,10 @@ this.playVideo = function( src ) {
     playVideo( id );
 }
 
+this.enableBlocklyTab = function ( objectID ) {
+    var tabs = this.blocklyTabs;
+    tabs.push( objectID );
+    this.blocklyTabs = tabs;
+}
+
 //@ sourceURL=source/scene.js
--- a/source/scene.js
+++ b/source/scene.js
@@ -394,11 +394,4 @@ this.playVideo = function( src ) {
     playVideo( id );
 }
 
-this.enableBlocklyTab = function ( nodeID ) {
-    var node = this.findByID( this, nodeID );
-    if ( node ) {
-        node.UIEnabled = true;
-    }
-}
-
 //@ sourceURL=source/scene.js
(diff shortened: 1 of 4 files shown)


BrettASwift:
My first thought would be to add a `UIEnabled` property (or whatever name you think is best) to our blockly nodes and when that property is set, add the blockly tab associated with the `nodeID`. We can add it as a behavior or make our blockly nodes extend a custom component that itself extends `blocklyController`.

AmbientOSX:
Where have I gone astray? (I promise I'll learn eventually :P)

source/scene.js
```diff
@@ -394,4 +394,11 @@ this.playVideo = function( src ) {
     playVideo( id );
 }
 
+this.enableBlocklyTab = function ( nodeID ) {
+    var node = this.findByID( this, nodeID );
```
BrettASwift: This is a costly find. I'd suggest either passing the node or the node name instead.

source/view/index.js
```diff
@@ -353,6 +355,20 @@ vwf_view.satProperty = function( nodeID, propertyName, propertyValue ) {
                 updateBlocklyRamBar();
                 break;
 
+            case "UIEnabled":
```
BrettASwift: The tabs are already generated when the node is initialized (check `vwf_view.initializedNode`). Here, just check `propertyValue`. If it's `true`, call `addBlocklyTab( nodeID )`, otherwise call `removeBlocklyTab( nodeID )`.

source/view/index.js
```diff
@@ -279,12 +279,14 @@ vwf_view.createdNode = function( nodeID, childID, childExtendsID, childImplement
             "ID": childID, 
             "name": childName,
             "ram": 15, 
-            "ramMax": 15
+            "ramMax": 15,
+            "UIEnabled": false
```
BrettASwift: Doesn't look like you're using this. I don't think it's necessary.

source/view/index.js
```diff
@@ -279,12 +279,14 @@ vwf_view.createdNode = function( nodeID, childID, childExtendsID, childImplement
...
         };
     } else if ( isGraphObject( protos ) && childName === "blocklyLine" ) {
         graphLines[ childName ] = { 
             "ID": childID, 
-            "name": childName
+            "name": childName,
+            "UIEnabled": false
```
BrettASwift: Same here.

source/view/index.js
```diff
@@ -381,6 +397,14 @@ vwf_view.satProperty = function( nodeID, propertyName, propertyValue ) {
...
             Blockly.SOUNDS_ = {};
             selectBlocklyTab( propertyValue );
+        }
+        else if ( propertyName === "blocklyTabs" ) {
+            clearBlocklyTabs();
+            var tabs = propertyValue;
+            var arrayLength = tabs.length;
+            for (var i = 0; i < arrayLength; i++) {
+                addBlocklyTab( tabs[ i ] );
+            }
```
BrettASwift: This can be removed.

AmbientOSX:
As a quick update - this somehow now breaks the second Blockly tab (on 2a for example). Looking into it...

BrettASwift:
Any update? I haven't looked into it yet.

AmbientOSX:
Closing this - assuming this functionality is moving to the HUD driver

PR #490, opened 2015-02-05
Dummy scenario with multiple rovers

nmarshak1337:
Reopening pull request - now the dummy scenario is in the Cartesian plane, has unnecessary triggers removed, has all blocks that we've previously used in the game implemented, incorporates math from the graph tool, and has an anomaly located where the anomaly in scenario3c is.


kadst43:
@AmbientOSX: Could you test?

AmbientOSX:
So Im a bit unclear what exactly a successful test case is here but all the rovers seem to be there and functioning independently.

nmarshak1337:
What I tested (someone else can feel free to reproduce) was:

1. See if all three rovers are there and functioning independently.
2. See if the graph tool works.
3. See if all Blockly blocks in the toolbox work. I tested some of the math (i.e. +, -, /, *) by stuffing into a variable.
4. Tested to see if the sensor works by placing an anomaly at [ 12, 10 ]. You can program Rover 1 to follow the line, then stop at the anomaly by following one of the programs in the cheat sheet (in particular, the one where the slope is 1/5... it can be found around Page 4 of the cheatsheet).

AmbientOSX:
In that case - those things seem to be working very well :)

source/rover.js
```diff
@@ -17,6 +17,15 @@ this.initialize = function() {
...
     }
+
+    //TODO: Set this properly (we shouldn't have to set this explicity in here...
+    // this.transform gets set automatically for ONE of the rovers, but not the other one)
+    this.transform = [
+        1,  0,  0,  0,
+        0,  1,  0,  0,
+        0,  0,  1,  0,
+        0,  0,  0,  1
+    ];
```
BrettASwift: I'm not sure I understand what's going on here. Why wouldn't the other rovers have a transform? What were you seeing prior to adding this code?

@eric79 Any thoughts? If there is something wrong here, it could be a larger issue.

source/scenario/scenario_dummy.vwf.yaml
```diff
@@ -0,0 +1,221 @@
...
+    - visible
+    - true
+  - setProperty:
+    - rover2
+    - visible
+    - true
+  - setProperty:
+    - rover
+    - battery
+    - 9001
```
BrettASwift: Vegeta would be impressed.

source/scenario/scenario_dummy.vwf.yaml
```diff
@@ -0,0 +1,221 @@
...
+  - setProperty:
+    - rover
+    - batteryMax
+    - 9001
+  - callMethod:
+    - rover
+    - allowedBlocksChanged
```
BrettASwift: This is an event that will be fired by the setting of `blockly_allowedBlocks`. It should not be done explicitly.

source/scene.js
```diff
@@ -146,7 +146,8 @@ function calcGridBounds( grid ) {
 this.executeBlock = function ( block, action ) {
     var blockName = block[ 0 ];
     var blockID = block[ 1 ];
-    this.blockExecuted( blockName, blockID );
+    //SJF: Was breaking procedures
+    // this.blockExecuted( blockName, blockID );
```
BrettASwift: I have a feeling that this is going to break something else. Blockly tracing?

nmarshak1337: Correct, this will break tracing for now. Possible next steps are to fix tracing, or to disable tracing for the dummy scenario / scenarios that use procedures.

BrettASwift: If we intend to keep tracing, it will need to work for procedures.

@kadst43 Any change in thinking on the necessity/usefulness of blockly tracing?

source/rover.js
nmarshak1337: If I comment this out, I run into trouble in the `placeOnTerrain` function - in particular, around line 108 (where `this.transform` is accessed). Currently, the first rover listed in `index.vwf.yaml` gets `this.transform` assigned before `placeOnTerrain` is called. The next two rovers have an undefined `this.transform`.

I.e. if the rovers are listed in the order `rover`,  `rover2`, `rover3` in `index.vwf.yaml`, then `rover` has `this.transform` assigned before `placeOnTerrain` is called, but `rover3` and `rover2` have an undefined `this.transform`. If we rerarrange `index.vwf.yaml` such that the rovers are ordered as `rover2`, `rover`, `rover3`, then `rover2` gets a transform before placeOnTerrain is called, but `rover` and `rover3` have undefined transforms.

BrettASwift: That's very strange. I'll take a look.

BrettASwift: Ok, it looks like the transform property is undefined until a translation, rotation, or transform is set (after initialization). This should probably not be the case as I'm sure the threejs object has a transform. I checked node3.vwf and there is no getter for the transform property. The others us an identity transform if the property has no value. Is there any reason transform should not do the same? @eric79 @davideaster

davideaster: Makes sense: https://github.com/virtual-world-framework/vwf/pull/447.

nmarshak1337:
Reopening pull request - now tracing works everywhere BESIDES the dummy scenario.

index.vwf.yaml
```diff
@@ -333,6 +333,243 @@ children:
...
+                    extends: http://vwf.example.com/node3.vwf
+                    children: 
+                      material:
+                        extends: http://vwf.example.com/material.vwf
+          cargo:
+            extends: source/inventory.vwf
+            properties:
+              capacity: 4
+              inventoryIsVisible: false
+
```
kadst43: I don't know that we need to do this right now - but we really ought to move the rovers into their own files at some point (rather than having all of this buried in the scene).

kadst43:
Looks good to me.

@BrettASwift?  Any final thoughts?

source/scene.js
```diff
@@ -27,6 +27,8 @@ var tiles = new Array();
 
 var lastCameraPOV = "thirdPerson";
 
+var scenario;
```
BrettASwift: We can't/shouldn't do this. Right now, I believe it still works due to an option in our config file. Component-scope private variables are no longer kosher in VWF. Correct me if I'm wrong, @eric79.

Where you need to access this, use the `scene.js` method `this.getCurrentScenario`.

source/scene.js
```diff
@@ -146,7 +148,11 @@ function calcGridBounds( grid ) {
 this.executeBlock = function ( block, action ) {
     var blockName = block[ 0 ];
     var blockID = block[ 1 ];
-    this.blockExecuted( blockName, blockID );
+    
+    if(scenario.name !== "scenario_dummy"){
```
BrettASwift: There are some coding standard violations here. Nothing serious, just some missing spaces. I sent an email about it. It may seem trivial, but it does help with code cleanliness and readability.

nmarshak1337: Cleaned up the missing spaces (I think?) Also made the comment more helpful.

BrettASwift:
:+1:

PR #491, opened 2015-02-05
Event fixes

AmbientOSX:
@eric79 Ignore the tab changes... Messed up a stash operation. Does everything here look kosher? I'm getting errors about 'scene' and 'this' being undefined now.

### DIFF
--- a/source/marsGameSound.js
+++ b/source/marsGameSound.js
@@ -20,15 +20,15 @@ this.initialize = function() {
     for ( var i = 0; i < soundSet.length; ++i ) {
         this.loadSound( soundSet[i] );
     }
-    this.future( 0 ).setUpSubtitles();
+    this.future( 0.01 ).setUpSubtitles();
 }
 
 this.setUpSubtitles = function() {
     var scene = this.find( "/" )[ 0 ];
     if ( scene && this.setUpOnce ) {
         this.setUpOnce = false;
-        this.soundStarted = this.events.add( startSubtitle, this );
-        this.soundFinished = this.events.add( stopSubtitle, this );
+        this.soundStarted = this.events.add( function( scene ) { startSubtitle(); }, scene );
+        this.soundFinished = this.events.add( function( scene ) { stopSubtitle(); }, scene );
     }
 }
 
--- a/source/marsGameSound.js
+++ b/source/marsGameSound.js
@@ -27,8 +27,8 @@ this.setUpSubtitles = function() {
     var scene = this.find( "/" )[ 0 ];
     if ( scene && this.setUpOnce ) {
         this.setUpOnce = false;
-        this.soundStarted = this.events.add( function() { startSubtitle.call( this ); }, this );
-        this.soundFinished = this.events.add( function() { stopSubtitle.call( this ); }, this );
+        this.soundStarted = this.events.add( function( instanceHandle ) { startSubtitle.call( this, instanceHandle ); }, this );
+        this.soundFinished = this.events.add( function( instanceHandle ) { stopSubtitle.call( this, instanceHandle ); }, this );
     }
 }
 
--- a/source/marsGameSound.js
+++ b/source/marsGameSound.js
@@ -27,8 +27,8 @@ this.setUpSubtitles = function() {
     var scene = this.find( "/" )[ 0 ];
     if ( scene && this.setUpOnce ) {
         this.setUpOnce = false;
-        this.soundStarted = this.events.add( function( scene ) { startSubtitle(); }, scene );
-        this.soundFinished = this.events.add( function( scene ) { stopSubtitle(); }, scene );
+        this.soundStarted = this.events.add( function() { startSubtitle(); }, this );
+        this.soundFinished = this.events.add( function() { stopSubtitle(); }, this );
     }
 }
 
--- a/source/marsGameSound.js
+++ b/source/marsGameSound.js
@@ -27,8 +27,8 @@ this.setUpSubtitles = function() {
     var scene = this.find( "/" )[ 0 ];
     if ( scene && this.setUpOnce ) {
         this.setUpOnce = false;
-        this.soundStarted = this.events.add( function() { startSubtitle(); }, this );
-        this.soundFinished = this.events.add( function() { stopSubtitle(); }, this );
+        this.soundStarted = this.events.add( function() { startSubtitle.call( this ); }, this );
+        this.soundFinished = this.events.add( function() { stopSubtitle.call( this ); }, this );
     }
 }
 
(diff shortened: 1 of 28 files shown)


source/marsGameSound.js
```diff
@@ -20,15 +20,15 @@ this.initialize = function() {
     for ( var i = 0; i < soundSet.length; ++i ) {
         this.loadSound( soundSet[i] );
     }
-    this.future( 0 ).setUpSubtitles();
+    this.future( 0.01 ).setUpSubtitles();
```
eric79: Why did the time need to change?

source/triggers/clauses/clause_OnBlocklyProgramChanged.js
```diff
@@ -60,11 +60,11 @@ this.onGenerated = function( params, generator, payload ) {
         }
 
         if ( this.addOrRemove !== "add" && object.blocklyBlockRemoved ) {
-            object.blocklyBlockRemoved = this.events.add( this.onProgramChangedEvent, 
+            object.blocklyBlockRemoved = this.events.add( function() { this.onProgramChangedEvent(); },
```
eric79: It looks like you missed the parameters for `onProgramChangedEvent`

source/triggers/clauses/clause_OnBlocklyWindowOpened.js
```diff
@@ -34,7 +34,7 @@ this.onGenerated = function( params, generator, payload ) {
         var object = blocklyObjects[ i ];
 
         if ( object.blocklyVisibleChanged ) {
-            object.blocklyVisibleChanged = this.events.add( this.onBlocklyEvent, 
+            object.blocklyVisibleChanged = this.events.add( function() { this.onBlocklyEvent(); },
```
eric79: I think that you missed the `visible` parameter here.

source/triggers/clauses/clause_OnScenarioStart.js
```diff
@@ -26,11 +26,11 @@ this.onGenerated = function( params, generator, payload ) {
     this.scenariosToCheck = this.extractStringArray( params[ 0 ] );
 
     // Setup the callbacks that should trigger us
-    this.scene.scenarioStarted = this.events.add( this.onScenarioEvent, this );
+    this.scene.scenarioStarted = this.events.add( function() { this.onScenarioEvent(); }, this );
```
eric79: I think that you are missing the `scenarioName` parameter here.

source/triggers/clauses/clause_OnScenarioSucceeded.js
```diff
@@ -26,11 +26,11 @@ this.onGenerated = function( params, generator, payload ) {
     this.scenariosToCheck = this.extractStringArray( params[ 0 ] );
 
     // Setup the callbacks that should trigger us
-    this.scene.scenarioSucceeded = this.events.add( this.onScenarioEvent, this );
+    this.scene.scenarioSucceeded = this.events.add( function() { this.onScenarioEvent(); }, this );
```
eric79: I think that you are missing the `scenarioName` parameter here.

source/triggers/clauses/clause_OnVideoPlayed.js
```diff
@@ -25,7 +25,7 @@ this.onGenerated = function( params, generator, payload ) {
 
     this.videosToCheck = this.extractStringArray( params[ 0 ] );
 
-    this.scene.videoPlayed = this.events.add( this.onPlayedEvent, this );
+    this.scene.videoPlayed = this.events.add( function() { this.onPlayedEvent(); }, this );
```
eric79: It looks like you are missing the 'video' parameter here.

eric79:
Looks like you're on the right path!  I just pointed out some places where it looked like there were missing parameters.

Do you know how to get those blockly tab commits out of this branch without losing them?  If not, let me know, and I can walk you through the git steps.

source/marsGameSound.js
kadst43: There was a conversation I had with... somebody.  @BrettASwift?  @davideaster?  Anyway, the sense I came away with was that a future( 0 ) wouldn't always get you past all the events you needed to wait for - you wanted to actually put it at a time in the future, so that all the things scheduled for the current time would truly be guaranteed to be done.

eric79: Okay, right, I remember that conversation.  It was basically that if your `application.initialize` is doing `future( 0 )` things that you need to happen before this, then you'll want to set this one up to run at `future( 0.01 )` so as to make sure that the `application.initialize` `future( 0 )` happens first.  I assume that is the case here?

kadst43: So my takeaway was that I should just always do future( something very small ) to be safe, since I don't have enough knowledge of VWF and its inner workings to make a better judgement call than that.  That said... this isn't my code, so I should shut up and let Spence (or whoever) answer now... :)

eric79: Okay, I think you would find that if you replaced every `future( 0 )` call with a `future( 0.01 )` call, you would be right back in the same situation you were before.  Basically, `future( 0 )` calls are placed on the queue in the order they are "scheduled" (when that line of code executes).  Since when the application is created, `initialize` calls are executed from the bottom up, children `future( 0 )` calls will all get scheduled before parent `future( 0 )` calls.  If a parent `future( 0 ) call needs to run first, then the child call should be scheduled for a later time (`future( 0.01)`).  This could get really complicated with a series of `future( 0 )`, `future( 0.01)`, `future( 0.02 )`, etc calls, but in practice we haven't run into that complicated of dependency structures.  If that ended up happening, there is probably a better way to do what needs to be done that should be investigated.

scottnc27603: Would an application loaded event solve this issue?

eric79: Yes, I think so.

AmbientOSX:
@eric79 For knowledge's sake - was hand-reverting that commit the appropriate way to do it?

BrettASwift:
@AmbientOSX You can rebase this branch on the one you intend to merge it with, select the "force rebase" option, and then right-click the commit you want to leave out and select "skip." This is assuming you're using the TortoiseGit GUI.

BrettASwift:
:+1: Let's get this into the App State branch.

eric79:
@AmbientOSX ...  I second what @BrettASwift said.  :smile:   It's better to remove the undesired commits entirely rather than have them there and then add a "reverting" commit (assuming that it hasn't already made it back to the main branch ... once it makes it to the main branch and there are more people working with it than just you, the "revert" would be the right way to do it so that you don't rewrite everyone else's history).

PR #492, opened 2015-02-12
Trigger vwfication

kadst43:
All the bugs aren't out yet, but this is close enough that you can test it to see if actions replicate properly (though I still need to do start state).


kadst43:
@AmbientOSX @BrettASwift

PR #493, opened 2015-02-12
Jplayer integration

nmarshak1337:
The jPlayer driver now lets us take URLs that point to files. Plus, we generate a videoEnded event when the video ends. Finally, we can show and hide the player, as well as call clearMedia.


kadst43:
@BrettASwift

index.vwf.html
```diff
@@ -30,11 +30,24 @@
...
+    <!--container in which our video will be played-->
+    <div id="jquery_jplayer_1" class="jp-jplayer"></div>
+ 
+    <!--main containers for our controls-->
+    <div class="jp-gui">
+        <div class="jp-interface">
+        </div>
+    </div> 
+</div>        
 <div id="version"></div>
```
nmarshak1337: @scottnc27603 Right now there's a dependency on the correct divs in index.vwf.html. This is also true of TDG, however, which uses the current, upstream jPlayer driver. I do not see this as a regression.

scottnc27603: I'm sure in both cases this was done to save time.  As long as it's not on
development then I'm ok with it.  It just needs to be turned into a string
with those ids externalized to the options.  That way the driver can insert
the html at the bottom of the body.

BrettASwift:
Can we close this since we're planning to merge it into App-State? Or, should we do JPlayer separate so we can fix the video bug if App-State requires more time to get running smoothly? @kadst43 @nmarshak1337

kadst43:
Yes, let's close this PR and get it mergfed into Application-State instead.

PR #494, opened 2015-02-14
Trigger vwfication

kadst43:
This should be ready to merge into Application-State, so that it can get hammered on with the rest of the replication fixes.

@BrettASwift @AmbientOSX


BrettASwift:
I want to test this before I give it the thumbs up. I should get to it tomorrow.

AmbientOSX:
Currently trying to debug an issue with scenario2b and the blockly graph tab/blocks not appearing.

source/triggers/actions/action_enableBlocklyTabs.js
```diff
@@ -0,0 +1,40 @@
...
+                            "the name(s) of the tab(s) to be enabled." );
+        return false;
+    }
+
+    this.tabNames = params;
+    return true;
+}
+
+this.executeAction = function() {
+    this.scene.clearBlocklyTabs();
```
AmbientOSX: So the current code asserts it cant find the tab but the following code works/fixes it:

var object;
    this.scene.clearBlocklyTabs();
    for ( var i = 0; i < this.tabNames.length; i++ ) {
        object = this.scene.find( "//" + this.tabNames[ i ] )[ 0 ];
        if ( object ) {
            this.scene.enableBlocklyTab( object.id );
        }
    }

AmbientOSX:
Aside from the 1 fix I mentioned this looks :+1: - seems like it will need a rebase as well

source/triggers/actions/action_enableBlocklyTabs.js
kadst43: This does exactly what my code does, except it doesn't print out the error to let you know that you tried to enable a blockly tab that doesn't exist.  Personally, I think we should print the error to let the designer know that they asked it to do something that doesn't make sense, rather than failing silently.

AmbientOSX: I'm confused - the implementation currently asserts that the tab is not found and makes the graph tab inaccessible in 2b... even though the tab is being found... this might be exposing an issue between this.scene.find and this.findInScene.

PR #499, opened 2015-02-18
Hud driver conversion

BrettASwift:

This is complete. All of the HUD elements have been converted, except for the Blockly trace element which we discussed.

Requires: https://github.com/virtual-world-framework/vwf/pull/466


source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
+        alpha = 0.5;
+      } else if ( element.isBlinking ) {
+        if ( time  - element.lastBlinkTime > element.blinkInterval ) {
+          alpha = 0.5;
+          if ( time - element.lastBlinkTime > element.blinkInterval + element.blinkDuration ) {
+            element.lastBlinkTime = time;
+          }
+        }
+      }
+      context.globalAlpha = alpha;
```
kadst43: Is there a reason to have the code in yaml?  Wherever possible, I'd rather have it in a .js file...

source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
+  elementPostDraw:
+    parameters:
+      - context
+      - element
+    body: |
+      context.globalAlpha = 1;
```
kadst43: Is this not using the element parameter?

source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
+  setAllBlinking:
+  setAllEnabled:
+children:
+  blocklyButton:
```
kadst43: Given that the children are moderately complex, would it make sense to give each child its own .js and .yaml files, maybe all in some subdirectory like source/hud (so we don't clutter up the source directory too much)?  I don't feel strongly for the HUD, but it seems like that might be a bit cleaner.

source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
+      blinkDuration: 0.25
+    methods:
+      draw:
+        parameters:
+          - context
+          - position
+        body: |
+          if ( this.icon ) {
+            context.drawImage( this.icon, position.x, position.y );
+          }
```
kadst43: Code in a .js file?

source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
```
kadst43: code in .js?

source/marsGameHud.vwf.yaml
```diff
@@ -0,0 +1,468 @@
...
+            case "topDown":
+              src = "assets/images/hud/camera_" + value + ".png";
+              break;
+            default:
+              src = "";
+              break;
+          }
+          var images = this.images;
+          images.activeModeIcon = src;
+          this.images = images;
```
kadst43: With a setter this complex, I would add a method named something like setActiveMode$, define it in the .js, and then just call it here.

source/marsGameHud.vwf.yaml
BrettASwift: Sort of. It helps when writing the draw functions to be able to see all of the properties available to you. The draw functions, by the way, are used only by the HUD driver. They're not true methods. I would like to find a way to make that clear, but I have not yet found a way. `this` in the draw function refers to the view object (HUD element) rather than the VWF node itself.

source/marsGameHud.vwf.yaml
BrettASwift: Nope. That's just there because the element is being passed in under the covers. It's more for our sake, just to know it's there if we need it. As per my other comment, I need to find a way to make some things more clear to the developer.

source/marsGameHud.vwf.yaml
BrettASwift: I suppose it won't be too much trouble.

BrettASwift:
I'll break up everything into their own files and push it up.

BrettASwift:
Moved the elements to their own files, added license info, etc. Had to update the driver to handle moving the draw functions into external scripts as well.

source/scenario/scenario1a.vwf.yaml
```diff
@@ -103,15 +103,15 @@ children:
             - [ 13, 15, -15 ]
           - setObjective:
             - "Click the PROGRAM button to open your control processor."
-          - setHUDProperty:
-            - camera_topDown
```
kadst43: Can we get rid of the setHUDProperty action?  Any other actions we can get rid of/simplify now?

kadst43:
I like this a lot better - much cleaner. :)

Aside from the one comment (and I assume that you've thoroughly tested... ;) ): :+1:

PR #500, opened 2015-02-25
Mars-Game Development Core Changes

BrettASwift:

This brings mars-game-development up to the latest VWF development. Most of this is relative path changes and prototype ID changes. This also disables trigger groups from checking while the app is not in the `playing` state.

My main concern here is performance. I had to turn down the trigger group frequency because it is taking a lot of processing power to check the groups. Everything appears to work, but slowly. There's more that can be done to make trigger group checking more selective so it's only checking triggers that are a part of active scenarios.

@davideaster We are currently using a future call to repeatedly fire the `checkTrigger$` function. Right now it's a ~0.055 future call each time. If we used future 0, would it just occur on the next tick and not schedule what appears to be a backlog of checks when the app is running slowly?

This branch works with: https://github.com/virtual-world-framework/vwf/tree/branch/mars-game-development

### DIFF
--- a/source/triggers/groups/triggerGroup.js
+++ b/source/triggers/groups/triggerGroup.js
@@ -16,7 +16,7 @@ this.initialize = function() {
     this.triggers = [];
     this.canFire$ = [];
     this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
-    this.future( this.checkFrequency$ ).checkTriggers$();
+    // this.future( this.checkFrequency$ ).checkTriggers$();
 }
 
 this.addTrigger = function( trigger ) {
@@ -58,7 +58,7 @@ this.checkTriggers$ = function() {
     //  this, do another check after a future(0).
     if ( haveTriggerToFire ) {
         this.future( 0.01 ).checkTriggersCallback$();
-    } else  {
+    } else if ( this.isChecking$ ) {
         // schedule the next check
         this.future( this.checkFrequency$ ).checkTriggers$();
     }
--- a/source/triggers/groups/triggerGroup.js
+++ b/source/triggers/groups/triggerGroup.js
@@ -42,7 +42,7 @@ this.checkTriggers$ = function() {
     var haveTriggerToFire = false;
     for ( var i = 0; i < this.triggers.length; ++i ) {
         var trigger = this.triggers[ i ];
-        this.canFire$[ i ] = trigger.isEnabled && trigger.check();
+        this.canFire$[ i ] = trigger.check();
         haveTriggerToFire = haveTriggerToFire || this.canFire$[ i ];
     }
 
--- a/source/triggers/groups/triggerGroup.js
+++ b/source/triggers/groups/triggerGroup.js
@@ -15,7 +15,7 @@
 this.initialize = function() {
     this.triggers = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
+    this.checkFrequency$ = 0.08 + ( Math.random() * 0.02 );
     // this.future( this.checkFrequency$ ).checkTriggers$();
 }
 
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -591,7 +591,7 @@ children:
             extends: source/inventory.vwf
             properties:
               capacity: 4
-              inventoryIsVisible: false               
+              inventoryIsVisible: false
   pickups:
     extends: http://vwf.example.com/node3.vwf
     children:
--- a/source/blocklyObjectGetter.vwf.yaml
+++ b/source/blocklyObjectGetter.vwf.yaml
@@ -25,5 +25,5 @@ methods:
   getBlocklyObjects:
 
 scripts:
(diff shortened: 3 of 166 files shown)


source/triggers/groups/triggerGroup.js
```diff
@@ -15,8 +15,8 @@
 this.initialize = function() {
     this.triggers = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
-    this.future( this.checkFrequency$ ).checkTriggers$();
+    this.checkFrequency$ = 0.08 + ( Math.random() * 0.02 );
+    // this.future( this.checkFrequency$ ).checkTriggers$();
```
kadst43: It is extremely mysterious to me why these would be causing a performance issue.  We have one trigger group with maybe a couple dozen triggers in it.  None of the checks that those triggers are doing are expensive.  This is maybe .01% of the amount of work that, say, rendering has to do - so what is making it so slow?

That said, this should be safe to check in - but eventually, we're going to need to understand why this is expensive and address it - it shouldn't even pop up on the radar.

kadst43:
Brett, there shouldn't be any way for it to schedule a backlog of checks, because each check schedules the next check - so until each check happens, the next one won't be scheduled. 

So, for example, on init the first check is scheduled to happen in 0.1 seconds (or however long - that's approximately right).  Then everything stops and nothing is processed for 10 seconds.  No more checks will be scheduled, because they're only scheduled inside of that call that is waiting 10 seconds to happen - and they're scheduled in a future that will push out from whatever time it is when the call actually occurs.

At least, that's how I would expect it to be working...

In any case - I'm more than a little bit concerned that the trigger system is causing performance issues - we'll want to understand why at some point, because the amount of actual work being done is very, very small.  With that said, this checkin looks safe, and if it patches the issues for the moment then that's a good first step.

Where you would see issues is on the success and fail conditions.  As you work through the scenarios, watch carefully for situations where you get more than one success/failure happening as a result of completing the same mission - if that's happening, we need to tweak the settings some more.

BrettASwift:
I was thinking that maybe because the time between ticks was greater than that future call then maybe it is detecting that the future call was supposed to be the time of the first call + future time. If that's the case, then it may be trying to make up for lost time, if you will. I could be completely wrong. If so, I'm not sure what is causing it to call that method so many times between renders and why removing it fixes the issue.

davideaster:
@kadst43's description is correct. I would add that these calculations happen on a simulation timeline, and that future actions run at exactly the time they're scheduled (in the pretend world of the simulation).

If an action is scheduled at 0.5 seconds, and it schedules another action for 10 seconds later, kernel time will be exactly 0.5 during the first action and exactly 10.5 for the second. The same is true for kadst43's example of 0.1 seconds, except that 0.1 has a rounding error in binary. Still, the actions execute at whatever 0.1 and 0.1 + 10 resolve to.

Simulation time doesn't match real time exactly, of course. It repeatedly lags behind and catches up as it receives authorization from the reflector to move forward, and as the client takes time to execute the actions.

The interval between tick messages doesn't really affect the total amount of work done. Unless you're catching ticks directly (which we're trying to get away from model-side), the times on the tick messages don't affect the calculation at all. We could tick at 100/sec or 1/sec and the same sequence of actions would run at the same times (in the pretend world). The only difference would be more jitter against real time in the slower case.

I tried restoring this future call and didn't seem to get a large number of trigger calculations. What should I be seeing? Is it apparent just after starting a new game?

BrettASwift:
Yes, the commented out future call would only cause it to fire once (it won't start firing consistently until "new game" is clicked). Lowering the `checkFrequency$` to its original value and then starting a new game should be fairly unplayable, depending on your machine.

kadst43:
There are some places I'm using real-world time.  It sounds like we should change those to use kernel time - that would address my worries about triggers not firing if the frame rate gets too slow.  @davideaster, how do I get query kernel time?  And are the units seconds?

davideaster:
From a component, it's `this.time`. From a driver, use `this.kernel.time()`. It's seconds, yes.

Also, `triggerGroup.vwf` is using `Math.random`. For replication to work, this should use the replicated random at `this.random`. It has the same `[0,1)` range as `Math.random`.

davideaster:
I have a fairly decent machine, so the `checkTrigger$` calls probably aren't as severe for me. I'll see if I can get a handle on how much work it's doing.

BrettASwift:
@AmbientOSX I added a fix to your changes. Scenario1a wasn't working because the parent block was `null` and `null !== undefined`. I made the conditional more general to account for that. If I missed something, let me know.

Tested this. Everything works. Merging. We need to use `vwf/branch/mars-game-development` from now on.

AmbientOSX:
@BrettASwift Yeah I noticed that while we were in the meeting and was trying to get in a fix but my internet dropped out >_< thanks

kadst43:
Ugh.  Already ran into that problem somewhere else in the trigger system as well.  Might want to do a find-in-files in the triggers for "undefined", and see if there are any more likely trouble spots...

PR #501, opened 2015-02-26
Blockly Procedure Tracing

AmbientOSX:
@kadst43 @BrettASwift Adds support for procedure tracing. Ended up being a bit harder than I anticipated because the procedure calls arent executed like normal blocks. I have to check whether the parent of a block called is a procedure and even then that only gives me the procedure definition block.


kadst43:
Looks reasonable to me.  If you can get it in tonight, go ahead and merge.

PR #503, opened 2015-03-04
Use VWF Date and Time 

AmbientOSX:
Small update to use random and time properties provided by VWF for replication safety @kadst43


nmarshak1337:
Looks good to me... I think I'll defer to someone else for the "official" thumbs-up.

AmbientOSX:
Oh it is noteworthy that I am using properties rather than functions as per the following comment by David:

“From a component, it's this.time. From a driver, use this.kernel.time(). It's seconds, yes.
Also, triggerGroup.vwf is using Math.random. For replication to work, this should use the replicated random at this.random. It has the same [0,1) range as Math.random.”

Can see his comment at: https://github.com/virtual-world-framework/mars-game/pull/500 a ways down.

Let me know if I was interpreting what you said a bit too literally @davideaster :)

davideaster:
Nope, this looks good. `time` and `random` are exposed to components as properties since they don't take any parameters.

PR #505, opened 2015-03-05
Midnight blue scout rover

nmarshak1337:
Default color of the scout rover is now midnight blue.


BrettASwift:
If the wheel digging into the ground issue is fixed, then :+1:

nmarshak1337:
@BrettASwift I think the "wheel digging into the ground" is an issue with the geometry in the level - I think the wheel is being clipped by the mini rover track. This only appears to happen at that particular location. 

image

(Note that the image features the old scout rover). 

What would be the best way to get started on this? I could either:

a) Look at the level geometry.
b) See if z-indexing / draw order has anything to do with it.

BrettASwift:
You sent an email not long ago with the blue rover model's wheel going through the ground. If the origin has been fixed or if it's just an issue on slopes, then that's fine. The above is something I'll be looking into shortly.

nmarshak1337:
It’s the exact same issue (either with the yellow rover or the blue rover). I checked the origin on the blue rover – IIRC it’s exactly the same as on the fixed yellow rover.


-        Nathan

BrettASwift:
Ok, sounds good. :+1: to merge.

PR #507, opened 2015-03-11
Multi-Rover HUD

BrettASwift:
This has turned into a much larger task than anticipated. It has exposed some issues in our code that I would like to address sooner rather than later, so I will be submitting PRs back to this branch with fixes. I wanted to get this up first so the changes here don't get lost in a huge PR down the road. I will not be merging this until I've fixed the necessary items in the following to do list.

TODO:
- [x] Rewrite `enableBlocklyTabs` action to a more general `enableBlocklyNodes` that will enable the tabs and rover icons and anything else that needs to be turned on or off.
- [x] Change how the Blockly window is displayed so it is no longer based on setting `blockly_activeNodeID`. The active Blockly node should be able to be set without opening the window.
- [Optional] Write Normal, Scout, and Heavy rover components that have a HUD icon property so rover selector icons are created when the rover nodes are created. That way, we don't have to hard code the rover selector HUD creation with knowledge of how many rovers might be in the scene.
- [Optional] Create a uniform style across HUD elements (with alpha). I have a few concepts together, but I'll leave this on the back burner until it's worth the trouble.

Requires: https://github.com/virtual-world-framework/vwf/pull/481


BrettASwift:
I'm doing the optional additional tasks in separate PRs. This is working and is ready for review/merge. Please review soon.

source/hud/roverSelector.js
```diff
@@ -0,0 +1,113 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+this.draw = function( context, position ) {
+    var rover, icon, posx, posy;
+    for ( var i = 0; i < this.rovers.length; i++ ) {
+        if ( this.rovers[ i ].enabled ) {
+            rover = this.rovers[ i ];
+            icon = this[ rover.id ];
```
AmbientOSX: Not an issue - just a learning question. This is used for accessing an object in the node hierarchy by it's ID?

AmbientOSX:
All good :+1:

source/hud/roverSelector.js
BrettASwift: Everything in the `draw` functions are actually in the hud scope (`this` refers to a hud element in the view, not the node). The method body is copied by the hud driver as well as all of the properties on the hud element nodes. So, the property `rovers` contains a list of objects ([defined here](https://github.com/virtual-world-framework/mars-game/blob/Multi-Rover-HUD/source/hud/roverSelector.js#L75)) that contain an `id` property that is the same as the image name stored on the hud element. The `images` property is not directly copied over to the hud element, rather the individual images belonging to that property are. So, this is just determining if the rover is enabled and then retrieving the icon to display if that's true.

I know it's a little confusing, but it's powerful enough to make it worth it.

PR #508, opened 2015-03-11
RAM grayout fix

nmarshak1337:
Moved the RAM grayout fix to mars-game.

### DIFF
--- a/source/hud/blocklyButton.js
+++ b/source/hud/blocklyButton.js
@@ -22,6 +22,14 @@ this.onClick = function() {
     var hud = this.parent;
     hud.lastBlocklyNodeID = hud.lastBlocklyNodeID || this.scene.player.rover.id;
     if ( hud.lastBlocklyNodeID !== undefined ) {
+        
+        var blocklyNode = blocklyNodes[ hud.lastBlocklyNodeID ];
+
+        if ( blocklyNode && Blockly.mainWorkspace ) {
+            Blockly.mainWorkspace.maxBlocks = blocklyNode.ramMax;
+        } else {
+            console.log("Either blocklyNode or Blockly.mainWorkspace are missing!");
+        }
         this.scene.blockly_activeNodeID = hud.lastBlocklyNodeID;
     }
 }
--- a/source/hud/blocklyButton.js
+++ b/source/hud/blocklyButton.js
@@ -22,14 +22,6 @@ this.onClick = function() {
     var hud = this.parent;
     hud.lastBlocklyNodeID = hud.lastBlocklyNodeID || this.scene.player.rover.id;
     if ( hud.lastBlocklyNodeID !== undefined ) {
-        
-        var blocklyNode = blocklyNodes[ hud.lastBlocklyNodeID ];
-
-        if ( blocklyNode && Blockly.mainWorkspace ) {
-            Blockly.mainWorkspace.maxBlocks = blocklyNode.ramMax;
-        } else {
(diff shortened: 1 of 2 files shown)


source/hud/blocklyButton.js
```diff
@@ -22,8 +22,16 @@ this.onClick = function() {
...
     hud.lastBlocklyNodeID = hud.lastBlocklyNodeID || this.scene.player.rover.id;
     if ( hud.lastBlocklyNodeID !== undefined ) {
+        
+        var blocklyNode = blocklyNodes[ hud.lastBlocklyNodeID ];
+
+        if ( blocklyNode && Blockly.mainWorkspace ) {
+            Blockly.mainWorkspace.maxBlocks = blocklyNode.ramMax;
+        } else {
+            console.log("Either blocklyNode or Blockly.mainWorkspace are missing!");
+        }
```
BrettASwift: This block of code should be in `index.js` under the `satProperty` handler for `blockly_activeNodeID`.

nmarshak1337: I think this is getting overridden by the `satProperty` in `blockly.js`? If I add the following to `index.js`:

```
case "blockly_activeNodeID":
    console.log("Active node ID sat in index.js!"); 
    break;
```

I never see the above message printed.

BrettASwift: That's because it's already being handled above that. Line 326.

nmarshak1337: I actually injected that test printout above Line 326, in the switch statement around [Line 293](https://github.com/virtual-world-framework/mars-game/blob/ram-grayout-fix/source/view/index.js#L293). 

I thought that it would have been caught before Line 326, but it isn't because `blocklyNode` always seems to be undefined when `satProperty` is called in `index.js`. Currently investigating...

BrettASwift: It is undefined when the property that the node belongs to is not a blockly node. That's why it did nothing for you. `blockly_activeNodeID` is a property of the scene, so the `nodeID` is the app id. It looks like you already fixed it, though.

source/view/index.js
```diff
@@ -529,6 +538,15 @@ function runBlockly() {
...
+        
+        var blocklyNode = blocklyNodes[ this.id ];
+
+        if ( Blockly.mainWorkspace ) {
+            Blockly.mainWorkspace.maxBlocks = blocklyNode.ramMax;
+        } else {
+            console.log("Blockly.mainWorkspace not found!");
+        }
+
         vwf_view.kernel.setProperty( appID, "blockly_activeNodeID", this.id );
```
BrettASwift: Since `blockly_activeNodeID` is being set here and `maxBlocks` is being set in `satProperty`, we should be able leave `setActiveBlocklyTab` as it was.

nmarshak1337:
@BrettASwift @kadst43 @AmbientOSX 

Brett and I are continuing to investigate the causes of this bug. It appears that we need to update `maxBlocks` before `loadToolbox` is called. See gist here:

https://gist.github.com/nmarshak1337/8d7f51c7848fb0195f16#file-gistfile1-txt

BrettASwift:
What confuses me is that by grabbing a block or switching tabs, the toolbox updates and the gray-out is gone. There must be some way to update the toolbox without having to make sure everything is done in perfect order or switching tabs/grabbing blocks. I'm looking into that now.

BrettASwift:
Found a potential solution, trying to find the right spot in code for it.

BrettASwift:
Added a fix in #511 @nmarshak1337

PR #510, opened 2015-03-12
Enable blockly nodes

BrettASwift:

- Changed `enableBlocklyTabs` to `enableBlocklyNodes` which enables the tabs and sets up the rover HUD icons.
- Improved enable/disable tabs by firing one event and passing an array of ids to the view, rather than firing an event for each tab.
- Improved the rover selector code by using only node IDs and removing rover node references.

Requires: https://github.com/virtual-world-framework/vwf/pull/482

### DIFF
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -69,6 +69,8 @@ methods:
   unpauseGame:
   resetHUDState:
   selectBlocklyNode:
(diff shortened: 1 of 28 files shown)


BrettASwift:
Due to time and availability constraints, I'm going to start merging these down.

PR #512, opened 2015-03-13
Fix Grid Collision

BrettASwift:
- Fixed the bug with rovers not updating their locations on the grid properly, causing collisions where the rover used to be.
- Modified some grid methods to use node IDs instead of nodes.
  - Tiles now store node IDs rather than node references in their objects arrays.
  - This eliminates most of the `valueJSFromKernel` warnings.


AmbientOSX:
Very cool :+1: Notice any performance gains with the ID change?

BrettASwift:
A little bit. It "gained responsiveness" a lot sooner than usual on my machine. Still a ways to go to make it smooth,

PR #483, opened 2015-01-16
DONT MERGE! Blockly proc and vars

AmbientOSX:
@kadst43 I needed to grab mars-game-staging and rebase it off of development to get this to work. Mars-game-staging has the assert capability but that hasnt made it into development. Whereas the components that make triggers/recursive node mapping work are only in development. ANYWAY! :)

The camera starts out in a funky spot but the entire procedure and the blocks are available to drag into the workspace from the toolbox. Mess around with it and I'll iterate on any feedback!

PR #513, opened 2015-03-18
Review Only: Trigger Properties Optimization

AmbientOSX:
HUD was being obscured by a div that is only removed when the video finishes playing. Video never plays, div is never removed. Per Brett's comment hide/show should be properties in the jPlayer driver, not methods. Seeing performance gains on my end. Double success bug still persists but replication is working with these changes. Requires: https://github.com/virtual-world-framework/vwf/pull/483

@kadst43 @BrettASwift


source/triggers/groups/triggerGroup.js
```diff
@@ -13,41 +13,41 @@
 // limitations under the License.
 
 this.initialize = function() {
-    this.triggers = [];
+    this.triggers$ = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.08 + ( Math.random() * 0.02 );
-    // this.future( this.checkFrequency$ ).checkTriggers$();
+    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
+    this.future( this.checkFrequency$ ).checkTriggers$();
```
BrettASwift: This line doesn't need to be uncommented (it can be removed entirely). This call will cause it to only run once on start up. At best it does nothing, at worst it adds a little bit of processing time during load. We start checking triggers consistently when the application state changes to `playing`.

kadst43: It's really easy to test that it's working (after you remove it) - if triggers work at all, you're good.

We might add a comment on the checkFrequency$ function, saying where it first gets called from, since that's now non-obvious.

source/scenario/mainMenu.vwf.yaml
```diff
@@ -17,7 +17,7 @@ extends: ../scenario/scenario.vwf
 properties:
   scenarioName: mainMenuScenario
 
-  nextScenarioPath: "introScreenScenario"
+  nextScenarioPath: "scenario1a"
```
kadst43: Ummmm... is this right?  More to the point, is this scenario even being used any more?  Maybe we should get rid of it completely?

source/scene.js
```diff
@@ -87,8 +89,9 @@ this.setApplicationState = function( state ) {
 }
 
 this.newGame = function() {
+    console.log('starting new game');
     this.applicationState = "playing";
-    this.activeScenarioPath = "introScreenScenario";
+    this.activeScenarioPath = "scenario1a";
```
kadst43: Does this not make it skip the intro movie?

source/scenario/mainMenu.vwf.yaml
AmbientOSX: I changed it when I was getting a strange behavior upon changing scene.newGame (that is - without clearing the cache it would still enter introScreenScenario and I thought this was the issue). Probably inconsequential but I'm actually not sure if we can remove the mainMenu yaml

BrettASwift: Yes, that scenario can be removed entirely.

kadst43:
Merging into development - it may not be "all the way" ready, but it seems like a good first step.

PR #514, opened 2015-03-19
Fix image case

BrettASwift:

This should fix the `camera_thirdPerson.png` bug on the Illinois server.


kadst43:
:+1:

PR #515, opened 2015-03-19
Blockly Check and Solve Blocks

AmbientOSX:
@kadst43 I'm sure there are some edge cases that still need solving. Notably we need to determine how we want to handle cases like 7/3... Everything is evaluated as if it were Javascript so some unexpected behaviors may occur. Let me know what you find and I will make changes.


kadst43:
This appears to be based off of mars-game-staging, not mars-game-development.  I attempted a rebase, but very nearly every file has a conflict, so I'm going to let you do it. :)

AmbientOSX:
Strange... Closing and opening another.

PR #516, opened 2015-03-19
Blockly Check and Solve

AmbientOSX:
@kadst43


kadst43:
There's still work to do (as per my rather long e-mail), but these changes seem good - go ahead and merge.

PR #517, opened 2015-03-20
Small improvements to execution framerate

AmbientOSX:
Reinstated the slower trigger updates and made a change to using vwf time in one location. @kadst43

### DIFF
--- a/heightmap_generator/three.js
+++ b/heightmap_generator/three.js
@@ -38,7 +38,7 @@ self.console = self.console || {
 
 		self.requestAnimationFrame = function ( callback ) {
 
-			var currTime = Date.now(), timeToCall = Math.max( 0, 16 - ( currTime - lastTime ) );
+			var currTime = this.time, timeToCall = Math.max( 0, 16 - ( currTime - lastTime ) );
 			var id = self.setTimeout( function() { callback( currTime + timeToCall ); }, timeToCall );
 			lastTime = currTime + timeToCall;
 			return id;
--- a/source/triggers/clauses/clauseProto_OnEvent.js
+++ b/source/triggers/clauses/clauseProto_OnEvent.js
@@ -71,7 +71,7 @@ this.reset = function() {
 
 this.evaluateClause = function() {
     var retVal = ( this.lastEventTime$ > 0 ) && 
-                 ( Date.now() - this.lastEventTime$ <= this.threshold$ );
+                 ( this.time - this.lastEventTime$ <= this.threshold$ );
     return retVal;
 }
 
--- a/source/triggers/groups/triggerGroup.js
+++ b/source/triggers/groups/triggerGroup.js
@@ -15,8 +15,8 @@
 this.initialize = function() {
     this.triggers$ = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
-    this.future( this.checkFrequency$ ).checkTriggers$();
+    this.checkFrequency$ = 0.08 + ( Math.random() * 0.02 );
+    //this.future( this.checkFrequency$ ).checkTriggers$();
 }
 
 this.addTrigger = function( trigger ) {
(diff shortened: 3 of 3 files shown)


source/triggers/groups/triggerGroup.js
```diff
@@ -15,8 +15,8 @@
 this.initialize = function() {
     this.triggers$ = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
-    this.future( this.checkFrequency$ ).checkTriggers$();
+    this.checkFrequency$ = 0.08 + ( Math.random() * 0.02 );
```
kadst43: Should this be Math.random, or something that replicates?

heightmap_generator/three.js
```diff
@@ -38,7 +38,7 @@ self.console = self.console || {
 
 		self.requestAnimationFrame = function ( callback ) {
 
-			var currTime = Date.now(), timeToCall = Math.max( 0, 16 - ( currTime - lastTime ) );
+			var currTime = this.time, timeToCall = Math.max( 0, 16 - ( currTime - lastTime ) );
```
kadst43: What is this line doing?  Doesn't seem obvious to me...  Time to call is going to be something between 0 and 16... is that seconds?  Milliseconds? ... after the last time that it was called?  Maybe clearer to do something like:

var nextCallTime = Math.max(currTime, lastTime + 16);

It would also be good to know what the difference between using Date.now() and this.time here.  Seems like it *should* be Date.now, since it's scheduling an event relative to simulation time, not world time.  If we use this.time, do we run the risk of this happening at different times on different systems?  Is that a problem?

source/triggers/clauses/clauseProto_OnEvent.js
```diff
@@ -71,7 +71,7 @@ this.reset = function() {
 
 this.evaluateClause = function() {
     var retVal = ( this.lastEventTime$ > 0 ) && 
-                 ( Date.now() - this.lastEventTime$ <= this.threshold$ );
+                 ( this.time - this.lastEventTime$ <= this.threshold$ );
```
kadst43: I thought that we needed to use Date.now() (a) to ensure that this replicates, and (b) to ensure that we don't have framerate issues.  If you use this.time and the framerate is very low, you might get outside of the threshold in which the event is considered "current" before it ever has a chance to fire.  I would not switch this back.

source/triggers/groups/triggerGroup.js
```diff
@@ -15,8 +15,8 @@
 this.initialize = function() {
     this.triggers$ = [];
     this.canFire$ = [];
-    this.checkFrequency$ = 0.04 + ( Math.random() * 0.02 );
-    this.future( this.checkFrequency$ ).checkTriggers$();
+    this.checkFrequency$ = 0.08 + ( this.random * 0.02 );
+    //this.future( this.checkFrequency$ ).checkTriggers$();
```
kadst43: Technically, it should be this.random * 0,04 if the base is going to be 0.08 - but that's just me being anal. :)

source/triggers/groups/triggerGroup.js
```diff
@@ -107,7 +107,7 @@ this.checkTriggersCallback$ = function() {
 
     // Math.random() returns [0, 1), so this will get us a random integer 
     //  in the range [0, numAtPriority).
-    var selectionValue = Math.floor( Math.random() * numAtPriority );
+    var selectionValue = Math.floor( this.random * numAtPriority );
```
kadst43: Yes.  I wouldn't expect replication to work otherwise.

source/triggers/clauses/clauseProto_OnEvent.js
kadst43: Oh wait... do I have those two functions backwards? 

We want to use replicated time here.  The same may be true of my comment above.

kadst43:
Go ahead and merge this - I know you're still fighting the random issue, but this helps a lot so let's get it in.

:+1:

PR #518, opened 2015-03-20
Remove VWF calls from the inner loop of trigger group checks.  

kadst43:
Saves maybe 2-3% of execution time, so not huge.

@BrettASwift


AmbientOSX:
:+1:

PR #521, opened 2015-03-23
HUD Redesign

BrettASwift:

Here's the new HUD. The pause menu was also recolored. I kept the green in the main menu for now.


kadst43:
:+1: 

Couple thoughts:

* I'm still not sold on this color of blue - but let's put it in for now and see how others like it.
* Whatever color we end up with, we should update the Blockly window to match (but that can be a separate change).

PR #522, opened 2015-03-24
Camera work

BrettASwift:

- Camera no longer snaps multiple times when changing view
- Camera no longer gets offset when switching from top down to another mode
- Camera now stays in the same camera view when switching between rovers
- Camera now properly stops following a target after being switched off of that target


kadst43:
:+1:

PR #524, opened 2015-03-29
Heavy rover integration

nmarshak1337:
Integrated heavy rover that is two tiles wide, plus generalized collision detection to objects that are more than one tile wide. The dummy scenario itself provides spaces where it can and can't move.


source/grid.js
```diff
@@ -153,22 +172,43 @@ this.addToGridFromWorld = function( object, worldCoord ) {
 }
 
 this.removeFromGrid = function( nodeID, gridCoord ) {
-    var objects = this.getObjectsAtCoord( gridCoord );
-    var index = objects.indexOf( nodeID );
-    if ( index !== -1 ) {
-        for ( var i = index; i < objects.length - 1; i++ ) {
-            objects[ i ] = objects[ i + 1 ];
+
+    var node = this.scene.findByID( this.scene, nodeID );
```
nmarshak1337: I've been told that find is slow... is there any easy way around this? These functions already take nodeIDs rather than pointers to nodes, so I kept it that way. But that means that we have to find the actual node by nodeID.

PR #527, opened 2015-03-31
Misc improvements

BrettASwift:
Fixed a few undefined errors in synchronization and rounded the battery count for the secondary rovers.

### DIFF
--- a/source/hud/roverSelector.js
+++ b/source/hud/roverSelector.js
@@ -161,18 +161,41 @@ this.setUpListeners = function( rover ) {
     rover.maxBattery = node.batteryMax;
     rover.ram = node.ram;
     rover.maxRam = node.ramMax;
-    node.batteryChanged = this.events.add( function( value ) {
+    node.batteryChanged = this.events.add( function( value, id ) {
+        var rovers = this.rovers;
+        var rover = this.getRoverByID( id );
         rover.battery = value;
+        this.rovers = rovers;
     }, this );
-    node.batteryMaxChanged = this.events.add( function( value ) {
+    node.batteryMaxChanged = this.events.add( function( value, id ) {
+        var rovers = this.rovers;
+        var rover = this.getRoverByID( id );
         rover.maxBattery = value;
+        this.rovers = rovers;
     }, this );
-    node.ramChanged = this.events.add( function( value ) {
+    node.ramChanged = this.events.add( function( value, id ) {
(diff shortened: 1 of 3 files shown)


source/hud/roverSelector.js
```diff
@@ -161,18 +161,41 @@ this.setUpListeners = function( rover ) {
     rover.maxBattery = node.batteryMax;
     rover.ram = node.ram;
     rover.maxRam = node.ramMax;
-    node.batteryChanged = this.events.add( function( value ) {
+    node.batteryChanged = this.events.add( function( value, id ) {
+        var rovers = this.rovers;
+        var rover = this.getRoverByID( id );
         rover.battery = value;
+        this.rovers = rovers;
     }, this );
```
kadst43: Is it valid to use "this" in an event callback?

source/hud/roverSelector.js
```diff
@@ -161,18 +161,41 @@ this.setUpListeners = function( rover ) {
...
+this.getRoverByID = function( nodeID ) {
+    var rover, i;
+    for ( i = 0; i < this.rovers.length; i++ ) {
+        if ( this.rovers[ i ].id === nodeID ) {
+            rover = this.rovers[ i ];
+            break;
+        }
+    }
+    return rover;
+}
```
kadst43: You consistently look up this.rovers before calling getRoverById, you might as well pass it in (and thus decrease the number of calls into this, which can be expensive (assuming that this is a VWF object).

Failing that, let's get in the habit of storing it in a local variable outside the loop, for the same reason.

source/hud/roverSelector.js
BrettASwift: That's why we pass `this` on the next line. Similarly, if you use a method as your callback for the listener, you'd be using `this`.

source/hud/roverSelector.js
BrettASwift: `this.rovers` is an array property with each item in the array containing a rover id and the battery and ram of that rover. I don't think it's that expensive. I can store it outside of the loop I suppose.

kadst43: Every call into this is expensive, because it has to be synchronized.  Moving similar calls out of loops was what got us the speedup that made the game playable.

kadst43:
See comments and do what you think makes sense... but :+1:

PR #528, opened 2015-03-31
Heavy rover integration

nmarshak1337:
Heavy rover plus relevant collision detection now works. Plus, rovers are in an area farther inside the Cartesian plane, where there are some interesting obstacles (i.e. there's a path that only the smaller rovers can go through).


nmarshak1337:
PR into the wrong branch T.T

PR #529, opened 2015-03-31
Heavy rover integration

nmarshak1337:
Heavy rover and relevant collision code added. Placed rovers near some interesting obstacles in Cartesian Plane.


source/grid.js
```diff
@@ -153,22 +172,43 @@ this.addToGridFromWorld = function( object, worldCoord ) {
 }
 
 this.removeFromGrid = function( nodeID, gridCoord ) {
-    var objects = this.getObjectsAtCoord( gridCoord );
-    var index = objects.indexOf( nodeID );
-    if ( index !== -1 ) {
-        for ( var i = index; i < objects.length - 1; i++ ) {
-            objects[ i ] = objects[ i + 1 ];
+
+    var node = this.scene.findByID( this.scene, nodeID );
```
nmarshak1337: I'd prefer not to use find() / findByID() if at all possible, since it's considered to be slow. I'm keeping the function definitions the same as before (take nodeID as input, and not a pointer to the node... other parts of the code assume this to be true). Is there any easy way around this? One possibility is to pass in a pointer to the object as an optional third argument.

kadst43: How often are you doing this?  It looks like something that happens very rarely, probably during scenario load.  Shouldn't be an issue there.  If you were doing it dozens or hundreds of times per frame, that would be different...

nmarshak1337: removeFromGrid() actually gets called every time we call moveObjectOnGrid(), so in theory, every time the rover moves. That's still much less often than every time per frame though.

source/grid.js
```diff
@@ -153,22 +172,43 @@ this.addToGridFromWorld = function( object, worldCoord ) {
...
+            if ( index !== -1 ) {
+                for ( var i = index; i < objects.length - 1; i++ ) {
+                    objects[ i ] = objects[ i + 1 ];
+                }
+                objects.length--;
+            } else {
+                this.logger.errorx( "removeFromGrid", "Unable to find tile with " +
+                    "ID: " + nodeID );
+                return false;
+            }
```
BrettASwift: This all worries me. It might be slow, for one, and if it doesn't look at the correct tiles, then we may leave behind IDs causing tiles to be blocked. I'm trying to think of a way that is quicker and safer. Did you test this a lot with multiple turns and verifying that the IDs are properly being removed?

nmarshak1337: I did test with multiple turns, but I'll go back and test it again, this time documenting exactly what I did to test it.

Right now speed can't be too bad since our heavy rover is 2x2. But it's an O(n^2) check, assuming the rovers are square. Is getObjectsAtCoord() something that might be slow, as well? Ultimately it does a lookup into this.tiles, and I remember Kevin saying that calls into "this" (assuming "this" is a vwf object) are expensive, since they need to be synchronized. 

One easy way to speed things up would be to consider the direction that the rover is moving in, and then only check "front" tiles, because those are the only ones that are going to collide with something else.

kadst43: Not sure why it would be slow.  You're looping over all of the tiles underneath the rover - there's not much way to avoid doing that, right?

In any case... let's get it checked in, then deal with issues if/when they come up.

:+1:

nmarshak1337:
Replaced with #537

PR #532, opened 2015-04-02
Scenario Timer

BrettASwift:

Here's the timer for the competition. It times each scenario individually and the aggregate time.


kadst43:
Looks awesome!  I didn't test scenario failure - I trust that you did (if you didn't, then please do).

If it's easy, let's get it merged into development too.

:+1:

PR #535, opened 2015-04-07
More Blockly Added Capability

AmbientOSX:
@kadst43 @BrettASwift @nmarshak1337 Adds categories, modifies a bunch of blocks connection code, adds others, demonstrates collapsing blocks, cleans out some unused old Block code, and fixes scrolling issues with Blockly UI changes.

Depends on: https://github.com/virtual-world-framework/vwf/pull/486

### DIFF
--- a/index.vwf.html
+++ b/index.vwf.html
@@ -113,16 +113,18 @@ limitations under the License.
 
 <!-- BLOCKLY TOOLBOX -->
 <xml id="toolbox" style="display: none">
-  <block type="rover_moveForward"></block>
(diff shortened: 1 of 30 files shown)


nmarshak1337:
FYI the indicator seems to point to the wrong place when using procedures. I also find one indicator (the procedure indicator?) "floating in space" sometimes, while the block indicator is doing its thing.

Finally if you drag blocks back into the toolbox, there's no way to get rid of them.

Apologies if these are all already known issues... @kadst43 mentioned something about the pointer earlier today.

kadst43:
:+1:

PR #536, opened 2015-04-07
Fix Tile Blinking

BrettASwift:

This is one of those really odd bugs that stresses the importance of making sure there are no errors in the console. The tiles not blinking were caused by an error being generated when `pov` was undefined in the `mountName` `satProperty` handler. This caused `mountName` to no longer be able to be set in the model, so it was stuck with a value of `default` after the camera pans to the radio. The tile blinking checks to make sure that the value of `mountName` is `topDown` or else it does not blink.


kadst43:
:+1:

PR #538, opened 2015-04-10
Blockly Radial Move & Blockly args fix

AmbientOSX:
@kadst43 @BrettASwift @nmarshak1337 

Depends on https://github.com/virtual-world-framework/vwf/pull/488

This is a work in progress since we're still figuring out collision handling. Should generally not be used in conjunction with other movement blocks for demo purposes.

### DIFF
--- a/source/rover.js
+++ b/source/rover.js
@@ -188,31 +188,41 @@ this.turnRight = function() {
     this.activateSensor( 'forward' );
 }
 
-this.checkRadialCollision = function( currentPosition, futurePosition, direction ) {
-    //origin: rover position
-    //direction: direction of movement
-    //near: far enough away from the rover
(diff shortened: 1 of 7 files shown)


kadst43:
I'm a bit reluctant to merge in the radial move until we have it at least safe to use...

kadst43:
If you haven't already, could you update the sandbox scenario so that a) Only Peregrine can use the radial move block, and b) Peregrine doesn't have the Forward or Turn blocks?

Then... test by moving Peregrine through Rosie's space both with and without collisions, and :+1:

kadst43:
Also, add a tooltip... "Moves the specified number of spaces along the X and Y axes."  Or something like that.

PR #539, opened 2015-04-17
Blockly Signal Block

AmbientOSX:
@kadst43 Also made some various changes to arithmetic blocks and number blocks that I assumed we'd want.

### DIFF
--- a/source/rover.js
+++ b/source/rover.js
@@ -141,9 +141,7 @@ this.placeOnTerrain = function( pos ) {
         this.transform[ 8 ], this.transform[ 9 ], this.transform[ 10 ], 0,
         pos[ 0 ],    pos[ 1 ],    pos[ 2 ],    1
     ] );
-
-    this.activateSensor( 'forward' );
-    this.activateSensor( 'signal' );
+    
 }
 
 this.translateOnTerrain = function( translation, duration, boundaryValue ) {
@@ -267,6 +265,8 @@ this.calcRam = function() {
 
 this.blockCountChanged = function( value ) {
     this.calcRam();
+    this.activateSensor( 'forward' );
+    this.activateSensor( 'signal' );
 }
 this.allowedBlocksChanged = function( value ) {
     this.calcRam();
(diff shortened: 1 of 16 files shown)


source/rover.js
```diff
@@ -314,6 +323,29 @@ this.activateSensor = function( sensor ) {
...
+
+        var deltaX = signalPos[ 0 ] - currentPos[ 0 ];
+        var deltaY = signalPos[ 1 ] - currentPos[ 1 ];
+
+        var radians = Math.atan2( deltaY, deltaX ); // In radians 
+        var heading = radians * ( 180 / Math.PI );
+
+        // Convert to 0 to 360 
+        if ( heading >= 0 ) {
+            heading = 180 - heading;
```
kadst43: What does this line do?  It seems wrong...

source/view/blocks.js
```diff
@@ -644,13 +645,68 @@ Blockly.Blocks[ 'controls_sensor_tracks' ] = {
...
+    this.setColour( 30 );
+    this.appendValueInput('INPUT')
+        .appendField('Base Signal °')
+        .setCheck(['OperatorAddSubtract','OperatorMultiplyDivide','LeftParenthesis','RightParenthesis','Conditional']);
+    this.setOutput(true, null);
+
+    var thisBlock = this;
+    this.setTooltip( function() {
+      var content = {
+        text: "Checks our scanner for base location in a 360° arc that starts on the X-axis. Not yet available on Perry!"
```
kadst43: Remove the "Not yet available on Perry!" part.

Also, on text like this let's call him Peregrine.  Perry is a nickname (that he hates) - I think that outside of us developers, Rosie may be the only one who uses it.

kadst43:
To test... set the signal in each of the four quadrants, then write a program that will take you to it using this block.

source/view/blocks.js
kadst43: Also, I might specify that it returns a value between 0 and 360 (or -180 and 180, or whatever).

kadst43:
After testing, and updating the tooltip, :+1:

kadst43:
@AmbientOSX: Where are we at on this PR?  Can we either close and re-open when it's ready, or merge and close?

AmbientOSX:
Its good to merge so.. I'll do so. Need it for new block value visibility changes I'll be working on tomorrow anyway :)

PR #540, opened 2015-04-21
Jplayer preload

nmarshak1337:
Modified `videoController.js` to support the changes made in [#489](https://github.com/virtual-world-framework/vwf/pull/489). Also added (commented out) code in `scenario_dummy` for testing of the preloading functionality. 

Depends on [#489](https://github.com/virtual-world-framework/vwf/pull/489).

### DIFF
--- a/source/scenario/scenario1b.vwf.yaml
+++ b/source/scenario/scenario1b.vwf.yaml
@@ -50,6 +50,15 @@ properties:
     - setHeading
     - 0
 
+  - callMethod:
+    - rover2
+    - setHeading
+    - 0
+
+  #- callMethod:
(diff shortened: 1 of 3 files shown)


kadst43:
Not sure what this PR is doing.  Looks like there are some setHeading calls for the rover - maybe because it needs a rebase?  Also, looks like most of the code that might actually do something is commented out.

Is there a corresponding VWF-side PR?

kadst43:
Ahhh, read the comment now.

This is fine to check in, I guess - but the job is only half done.  Each scenario should specify which cutscene to have preloaded, and then this functionality should be used to make that happen.

PR #541, opened 2015-04-23
All scenarios specify which video to preload / video load on URL set

nmarshak1337:
Video now loads when URL is set, as the "preload" option is specified in the videoManager. Also, each scenario knows which video to preload. 

EDIT: Depends on [#492](https://github.com/virtual-world-framework/vwf/pull/492)


nmarshak1337:
Now ( with commit eaaf038c6555e4e70dc243d67a7d21ee8269a417 ) we preload the first cutscene at the menu screen.

kadst43:
:+1:

PR #542, opened 2015-04-24
Blockly multi tracing and looping

AmbientOSX:
@kadst43


kadst43:
See e-mail... but :+1: for these changes.

PR #545, opened 2015-04-30
Blockly Refinements

AmbientOSX:
Changelog:
- Fast and slow button
- Collision block (2 tiles ahead for rosie)
- Fix heading block and signal block issues (math)
- Fix Perry tracing
- Number block that you can type in that is limited to 3 digits (integers only)
- Show variable values in variable block
- Round signal degrees to int (but do not truncate)
- Return -1 on the signal move block if signal is at current location
- Enable connecting a negative to a conditional (>)
- Number block that contains ordered pair (x, y)
- GetX/GetY block
- Fix Proximity block (Am I next to this tile? - Any tiles around the rover for manny only)
- Anomaly Adjacent
- Parens around ordered pairs
- Variable windows -> text
- Degree symbols

Depends on https://github.com/virtual-world-framework/vwf/pull/494

I'd imagine there will be a few bugs with which blocks can be connected, block math, initialization timing of variables/sensors. I'm still seeing some weird things with the anomaly block.

@kadst43 @BrettASwift


source/rover.js
```diff
@@ -378,10 +397,13 @@ this.calcRam = function() {
 
 this.blockCountChanged = function( value ) {
     this.calcRam();
-    this.activateSensor( 'forward' );
+    this.activateSensor( 'anomaly' );
     this.activateSensor( 'signal' );
     this.activateSensor( 'heading', this.heading );
+    this.activateSensor( 'collision' );
+    this.activateSensor( 'proximity' );
```
kadst43: I think we talked about this... let's make the proximity sensor and the anomaly sensor be one and the same?

kadst43: (looks like maybe you're already doing that elsewhere?  Ignore me if I'm confused :) )

kadst43:
:+1:

PR #546, opened 2015-05-02
Projected grid shader

BrettASwift:
Here's a basic version of the projected grid shader. I don't have an example in mars-game yet, but I do have a screenshot from a test application:

projected_grid

This is the mix map shader + the projected grid. There are several uniforms that affect the appearance of the grid (color, thickness, etc). I'm still modifying how these uniforms behave, so this isn't done yet. I just wanted to get a preview up.


kadst43:
Looks awesome. :grinning: 

:+1:

PR #549, opened 2015-05-06
OnBlocklyBlock - Rover Sounds - Exe Small Fix

AmbientOSX:
@kadst43 I cant figure out why the new OnBlocklyBlock trigger is getting ignored. Necessary to determine when a rover is idling. Everything else should be all set. Rovers are playing their own sounds.

### DIFF
--- a/source/scenario/globalScenarioTriggers.vwf.yaml
+++ b/source/scenario/globalScenarioTriggers.vwf.yaml
@@ -222,6 +222,80 @@ properties:
       actions:
       - stopSound:
         - roverSounds
+      - playSound:
+        - roverEndSounds
+
+    startRover2Sounds:
+      triggerCondition:
+      - onBlocklyStarted:
+        - rover2
+      actions:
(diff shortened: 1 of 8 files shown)


source/scenario/globalScenarioTriggers.vwf.yaml
```diff
@@ -222,6 +222,80 @@ properties:
...
+        - roverSounds
+
+    rover3DeIdle:
+      triggerCondition:
+      - onBlocklyBlock:
+        - rover3
+        - rover_moveForward
+      actions:
+      - playSound:
+        - rover3Sounds
```
kadst43: We should probably take all of this out of the trigger system.  There's no reason to be driving sounds with triggers.  Instead, let's start adding event handlers to trigger sounds like these inside of the sound manager.

kadst43:
Go ahead and merge, but we should circle back and move these out of the triggers at some point.

PR #550, opened 2015-05-08
Shader fixes

BrettASwift:
@kadst43 @AmbientOSX 

Fixed a couple of bugs and added the complete shader code for lights and shadows from threejs to all of our current custom shaders.


kadst43:
:+1:, but wait until after the sprint demo

PR #552, opened 2015-05-13
Triangle Blocks, Position Sensor, Not Block, Rover Name Changes, Drawing Triggers

AmbientOSX:
@kadst43 @BrettASwift Also includes a number of miscellaneous bug fixes. I'm testing the triggers now but I wanted to get some eyes on before days end. Also widens the Blockly UI which should be a welcome relief...


source/rover.js
```diff
@@ -509,7 +513,7 @@ this.activateSensor = function( sensor, value ) {
         var proposedNewGridSquare = [ this.currentGridSquare[ 0 ] + dirVector[ 0 ], 
                                                                 this.currentGridSquare[ 1 ] + dirVector[ 1 ] ];
         arrayToCheck.push( proposedNewGridSquare );
-        if ( this.name === 'rover2' ) {
+        if ( this.name === 'rover3' ) {
```
kadst43: I'm confused.  This looks like code for Perry, but rover3 is Rosie...

source/scenario/globalScenarioTriggers.vwf.yaml
```diff
@@ -274,7 +274,7 @@ properties:
       triggerCondition:
       - onBlocklyBlock:
         - rover3
-        - rover_turn
+        - 'Turn: Left'
```
kadst43: What about right turns?

kadst43: Also, we should still make sounds while the rover is turning - just not when it's executing blocks *other* than move or turn.  And we will likely have custom sounds for other blocks (like the spray-can sound that gets turned on when drawing triangles, perhaps sounds for sensors, ...)

kadst43:
Let's get this merged (I want to work on mission 4, and this is a prerequisite), but also see comments...

:+1:

PR #553, opened 2015-05-14
Triangles and triggers

AmbientOSX:
@kadst43 Closed last PR, this one does not have the scary bug.


source/scenario/scenario_dummy.vwf.yaml
```diff
@@ -166,6 +166,15 @@ children:
           - setObjective:
             - "Explore the Cartesian Plane!"
 
+        onTriangle:
+          triggerCondition:
+            - onBlocklyPolygon:
+              - rover2
+              - [ [21,12] , [22,13] , [22,12] , [21,12] ]
```
kadst43: So, a couple things...

First, for now at least let's assume that all polygons are closed.

Second, let's assume that we don't care either about which vertex they start on or about which direction they trace the vertices in - that is, they can start anywhere and go around the polygon either clockwise or counterclockwise.  If, somewhere down the road, we want to be more strict then we can add boolean parameters to restrict what we accept, but let's handle the most general case right now.

source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -0,0 +1,76 @@
...
+    if ( this.blockNode === blockNode ) {
+
+        var equal = true;
+        var forward = true;
+        var backward = true;
+
+        for ( var i = 0; i < this.pointArray.length; i++ ) {
+
+                var thisPoint = this.pointArray[ i ];
+                var thatPoint = pointArray[ i ];
```
kadst43: Nitpick: Let's call the argument something that's not the same as our local property.  Maybe arrayToCheck?

source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -0,0 +1,76 @@
...
+                if ( thisPoint[ 0 ] !== thatPoint[ 0 ] || thisPoint[ 1 ] !== thatPoint[ 1 ] ) { 
+                    forward = false;   
+                }           
+
+                thisPoint = this.pointArray[ i ];
+                thatPoint = pointArray[ pointArray.length - 1 - i ];
+
+                if ( thisPoint[ 0 ] !== thatPoint[ 0 ] || thisPoint[ 1 ] !== thatPoint[ 1 ] ) { 
+                    backward = false;   
+                }
```
kadst43: Hrm.  I don't think this will quite handle all cases.  Try starting from different vertices, not just reversing order...

My algorithm would the as follows:

1) Loop through their array to find the vertex that matches vertex 0 in your array.  Fail if you don't find it.
2) Check the vertices before and after the one you found, and see which one matches vertex 1 in your array.  Fail if neither does.  Remember to handle the case where you have to wrap around from the beginning to end of the array (or end to beginning).
3) From there, loop in the direction you just determined, matching up each of the remaining vertices.

Of course, this may not work if both vertex 1 and the last vertex in your array are the same - so at load time, print an error if that's the case.

kadst43:
Let's get this merged first, and then see comments. :)

:+1:

PR #554, opened 2015-05-14
Polygon Trigger Generalization

AmbientOSX:
@kadst43 Accounts for forward and reverse cyclical permutations and allows a user to start at any point in the polygon.

### DIFF
--- a/source/triggers/clauses/clause_OnBlocklyPolygon.js
+++ b/source/triggers/clauses/clause_OnBlocklyPolygon.js
@@ -36,38 +36,50 @@ this.onGenerated = function( params, generator, payload ) {
     return true;
 }
 
-this.onPolygonFinished = function( blockNode, pointArray ) {
+this.onPolygonFinished = function( blockNode, playerPoints ) {
+
+    // Note: The playerPoints array should have length v+1 for the correct polygon. This is because we mark
+    // the point where the polygon is started and where it finishes to determine if it is closed.
 
     if ( this.blockNode === blockNode ) {
 
-        var equal = true;
-        var forward = true;
-        var backward = true;
+        // Does the player end up where they started? ( Closed Polygon )
 
-        for ( var i = 0; i < this.pointArray.length; i++ ) {
+        var startPoint = playerPoints[ 0 ];
+        var endPoint = playerPoints[ playerPoints.length - 1 ];
 
-                var thisPoint = this.pointArray[ i ];
-                var thatPoint = pointArray[ i ];
-
-                if ( thisPoint[ 0 ] !== thatPoint[ 0 ] || thisPoint[ 1 ] !== thatPoint[ 1 ] ) { 
-                    forward = false;   
-                }           
-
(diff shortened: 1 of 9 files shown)


source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -36,38 +36,50 @@ this.onGenerated = function( params, generator, payload ) {
...
 
-        }
+        var forwardArray = this.pointArray;
+        var reversedArray = forwardArray.slice( 0 );
 
-        if ( this.pointArray.length !== pointArray.length ) {
-            equal = false;
-            return;
-        }
+        reversedArray.reverse();
```
kadst43: Nitpick: Do this at initialization time, and store it?  Or why not just loop over the array backwards instead of making all this work?

source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -36,38 +36,50 @@ this.onGenerated = function( params, generator, payload ) {
...
+
+        var concatenatedForward = forwardArray.concat( forwardArray );
+        var concatenatedReverse = reversedArray.concat( reversedArray );
```
kadst43: Again, do this once at init time?

kadst43: Or, instead of doubling the array, just set your index counter to indexCounter % arraysize.  That will work as long as it doesn't go below zero.

source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -36,38 +36,50 @@ this.onGenerated = function( params, generator, payload ) {
...
+
+        var forwardString = concatenatedForward.toString();
+        var reverseString = concatenatedReverse.toString();
+
+        // Remove last point in the player array since it should be a duplicate
+
+        playerPoints.pop();
```
kadst43: Better to leave it in place and ignore it - no reason for this function to have a side effect on the param passed in...

source/triggers/clauses/clause_OnBlocklyPolygon.js
```diff
@@ -36,38 +36,50 @@ this.onGenerated = function( params, generator, payload ) {
...
+
+        var playerString = playerPoints.toString();
 
-        if ( equal === true && ( backward === true || forward === true ) ) {
+        if ( forwardString.indexOf( playerString ) !== -1 || reverseString.indexOf( playerString ) !== -1 ) {
```
kadst43: Ahhh, I see.  This is kinda slick, yes.  I'd still generate the strings at init time, and store them as properties.

kadst43:
A bunch of nitpicks, then :+1:

AmbientOSX:
@kadst43 Made changes but also snuck in a bunch of things from the email.

kadst43:
:+1:

PR #555, opened 2015-05-14
New terrain

BrettASwift:
@kadst43 @AmbientOSX 

This is the new terrain with the mix map. I added the new heightmap, but not the grids for the new scenarios. That will have to be done in another PR.


kadst43:
:+1:

kadst43:
Merging so I can build the grid.

PR #557, opened 2015-05-15
New Rosie

BrettASwift:
Added new heavy rover assets and turned down the environment light intensity. It was washing out the HUD pretty significantly.

@kadst43 @AmbientOSX


kadst43:
:+1:

PR #560, opened 2015-05-15
Pod assets

BrettASwift:
@kadst43 @AmbientOSX 

Here are all of the new base assets that were on redmine. I have no idea what size they are, but I added the components to the source/buildings folder.


kadst43:
:+1:

PR #562, opened 2015-05-17
Triangle outline

BrettASwift:
@kadst43 @AmbientOSX 

This has the triangle stuff and some other shader improvements. The grid is fully functional now with axis lines and all, whether or not we use it at this point.


kadst43:
:+1:

PR #564, opened 2015-05-17
Added function to scene to draw the triangle

BrettASwift:
@kadst43 

Added the `drawSchematicTriangle` function to the scene. You can hook up to that in your trigger action.


kadst43:
Merging, will make changes after.

PR #569, opened 2015-05-18
remove supplies from mission 3

kadst43:


kadst43:
super simple - just merging

PR #580, opened 2015-05-20
Makes the delay clause actually do something!

kadst43:
Also makes the mission brief open after the dialog for 3-1 and 4-1.  I'm using a delay, which is the wrong
way... but I'll debug the soundFinished event next.


kadst43:
That... wasn't easy.

AmbientOSX:
:+1:

PR #581, opened 2015-05-20
Misc Fixes 5-20-15

AmbientOSX:
@kadst43 @BrettASwift Success + failure is still happening on the triangle missions but I wanted to get these other fixes in. Still not sure what is going on there.

...also quickly added a number block that lets you visually select an angle that corresponds with the angles we're using in the environment. Might help with the "showing with hands" issue described on the call. Use it or ignore it.

### DIFF
--- a/source/rover.js
+++ b/source/rover.js
@@ -773,4 +773,12 @@ this.setHeading = function( newHeading, duration ) {
     
 }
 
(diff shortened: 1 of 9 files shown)


kadst43:
:+1:

PR #584, opened 2015-05-28
Mission Brief Polish

BrettASwift:
@kadst43 @AmbientOSX 

- Added a HUD icon to bring up the mission brief.
- Moved mission brief info into the mission tasks themselves.
- Redesigned mission brief window.

The last thing that I didn't finish was allowing each task to specify an image to go on the left side of the window. The idea being that each task could display an image that pertains to the task at hand. For instance, in scenario1h, the image could be of the radio module. For now, it's the default image of the rover.


kadst43:
:+1:

PR #585, opened 2015-06-01
Misc bug fixes

BrettASwift:
@kadst43 @AmbientOSX 

Fixes for Blockly bugs in scenario 2 and a fix for checking `surroundingTiles`. Sometimes it was `null` and `null !== undefined` is true. Converted to a boolean to make the check more robust.


AmbientOSX:
:+1:

PR #587, opened 2015-06-03
Review Only: Solar panel building

AmbientOSX:
@BrettASwift @kadst43 I'm a bit stuck here for a few reasons. New terrain in the solar panel area is breaking ( or not breaking, depending on whether we consider Perry movement failing on areas with more severe gradient working as intended ) Perry movement. Also, more critically, the trigger to construct the base component ( solarPanel1 ) seems to be working but the solar panel shows up beneath the terrain.

Just requesting a review of the implementation ( object.construct() should be all I need in the trigger, right? ) and any thoughts on the new terrain ( whether we want to smooth it out or whether I should adjust Perry's raycast collision detection ) before I proceed further.


kadst43:
Perry should be able to move on slopes.

Depending on how you're placing the solar panel - if you use the add to grid action in the scenario, it should groundfit it, but if you're just placing it in index.vwf.yaml then you probably need to set the z position.

AmbientOSX:
Ok I will tweak the collision raycasting. Right - I use addToGrid in the scenario definition.

kadst43:
Ok, you probably need @BrettASwift then. :)

BrettASwift:
In `setHeightFromTerrain` in `grid.js`, either switch to get the height value from the heightmap or change the z value of the `origin` variable to be higher. It's probably starting the raycast beneath the ground.

AmbientOSX:
I think I made the necessary changes but it still seems to be off. The Cargo Pod is also showing up beneath the terrain...

AmbientOSX:
Merging to get the Perry fixes in and some of the bug fixes.

PR #590, opened 2015-06-17
Exorcism of the grid system

BrettASwift:
@kadst43 @AmbientOSX 

Completely removed the old grid and inventory system. There's a lot here and there may be some bugs. In this branch, mission 4 and beyond do not work, but that's because of the radial raycast. I need to merge this to the grid-tool branch and rebase to see if I can fix some of that. This will not go into mars-game-development until I have tested it thoroughly and fixed any remaining bugs. This one might be worth a meeting to explain all of the new pieces.


kadst43:
:+1:

PR #592, opened 2015-06-19
New rover sounds

jyucra:
I added into, engine, outro and gravel sounds to match each of the rovers. Changes were also made for a spraying noise to be played when setting down the nano goo in a triangle, and for the engines to keep running even when the rovers aren't moving forward, but those depend on onBlocklyBlock which isn't working at the moment, so they're commented out in the mean time.
A minor change was made so that the VO file for scenario1a starts playing as soon as the player switches to that scenario.
There were also two changes made to onBlocklyBlock to try and catch potential bugs.


AmbientOSX:
:+1: Test thoroughly and then merge using GitHub's 'Merge Pull Request' functionality.

kadst43:
I grabbed this branch, loaded the test scenario, and tried it.  I saw the following issues:

1) Rosie has Perry's move block, not Manny's blocks.
2) Neither Rosie nor Perry actually moved.

These may not be new, but let's not merge until we confirm that this problem isn't introduced by this branch.

kadst43:
What is the plan with this PR?  We should either merge it (with fixes if necessary) or close it out...

PR #594, opened 2015-06-19
Blockly tooltips 2

jyucra:
I changed the tooltips for the custom Blockly blocks so that they always appear in a fixed location in the bottom right corner of the screen. There were changes made to the size of some of the tooltip images in order to better accommodate them in a more restricted space.


AmbientOSX:
:+1: We were having some trouble with integrating the rebuilt Blockly code to get the tooltips to appear immediately but we'll investigate that this sprint. Like the other PR - test and then merge this in first thing Monday.

PR #595, opened 2015-06-22
Changed the way the rover's names show up on the blockly tabs.

jyucra:
Rover, Rover2, and Rover3 have been changed to Manny, Perry, and Rosie respectively within the Blockly tabs.


AmbientOSX:
:+1:

PR #596, opened 2015-06-23
VO change for hint about deleting blocks

jyucra:
Changed the VO and message for the hint in scenario1f to reflect the preferred way of deleting blockly blocks.


AmbientOSX:
:+1:

PR #597, opened 2015-06-28
Update tooltips to match UI.

BrettASwift:
Updated the tooltip window to look more like the logger box. Changed height restriction so that smaller tips have less negative space.

@kadst43 @AmbientOSX @jyucra


AmbientOSX:
:+1:

PR #599, opened 2015-06-30
Temporary mission three

jyucra:
This branch replaces the old Mission 3 with a new mission which follows directly from the "To Be Continued" animation. It also includes changes to Mission 4 that allow it to reflect the changes which occurred in Mission 3. In addition to this:
- The number and order of solar panels created is changed.Now the rover holds off construction until he's building the diamond consisting of 4 solar panels.
- The green blockly line will only be visible in tasks which include the graph node.
- The naming scheme has been updated, scenario 1 is now Mission 1, and scenarios 2 and 3 have been combined into Mission 2.
- The base pod has been temporarily made invisible, as it will no longer be in the same location it was previously.

Still to do: The failure/success triggers in Mission 4 still need to be tweaked.


AmbientOSX:
Probably want to change the filenames of the scenarios as well? And make sure there's nowhere else in the code we reference the old naming convention for scenarios (whether thats a default value in index.vwf.yaml or the names of the Blockly XML files). Though I guess that is ok for a separate PR since this is already quite a few changes.

 If you can play through this morning and everything continues to work then :+1:

PR #600, opened 2015-06-30
Unnecessary blocks cleanup

jyucra:
Small changes to the blocks available to the player in Mission 4 in order to make it less confusing. Also makes a change to the opening VO which was supposed to go in with a previous branch that ultimately wasn't merged.


AmbientOSX:
:+1:

PR #601, opened 2015-06-30
Further tooltip changes

jyucra:
Makes the tooltips disappear when the mouse hovers out of a specific blocks instead of disappearing whenever a mousemove event is triggered.


AmbientOSX:
:+1:

PR #602, opened 2015-07-01
Blockly fixes 6 30

AmbientOSX:
@jyucra @kadst43 @BrettASwift Contains blockly window in the window, hides the program counter, adds the drawTriangle block and fixes double success and failure.

Issue with double success and failure related to trigger groups - will dig into figure out why they aren't working correctly but for now the simple fix is to temporarily disable failonincompleteprogram when the success condition is fired. The fact that Blockly stops executing during the scenario transition is the culprit here.


kadst43:
- Change the triangle name to "Draw Triangle" (with a capital 'D' and a space).
- Change the argument names to "Corner #[1-3]"
- Remove the old program from all of the old mission 4 tasks

Then :+1:.

PR #606, opened 2015-07-05
Nanite Particles

BrettASwift:
@kadst43 @AmbientOSX 

This adds the nanite particle component. It's not hooked up yet. In order to use it, we need to create one `naniteParticle` system for each line we draw and then set the start and stop points to the vertices of the line we're drawing. We may also want to have the stop point be perry as he moves to the actual stop point and then set the stop point to that vertex at that time.


AmbientOSX:
:+1:

PR #608, opened 2015-07-08
Fix onBlocklyBlock & refactor drawing block code & change block content

AmbientOSX:
@jyucra @BrettASwift @kadst43 

Removes some log code that was left in as well...

### DIFF
--- a/source/view/blocks.js
+++ b/source/view/blocks.js
@@ -1619,13 +1619,13 @@ Blockly.Blocks['draw_triangle'] = {
         .appendField("Draw Triangle");
     this.appendValueInput("pointA")
         .setCheck(null)
-        .appendField("Corner #1");
+        .appendField("Vertex #1");
     this.appendValueInput("pointB")
         .setCheck(null)
-        .appendField("Corner #2");
+        .appendField("Vertex #2");
     this.appendValueInput("pointC")
         .setCheck(null)
-        .appendField("Corner #3");
+        .appendField("Vertex #3");
     this.setPreviousStatement(true);
     this.setNextStatement(true);
     this.setColour(315);
--- a/source/view/blocks.js
+++ b/source/view/blocks.js
@@ -1644,8 +1644,6 @@ Blockly.JavaScript['draw_triangle'] = function(block) {
   var op_b = op_b_str.split(",");
   var op_c = op_c_str.split(",");
 
-  console.log(op_a);
-
   var actionA = {
     nodeID: block.data,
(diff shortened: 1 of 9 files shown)


kadst43:
@BrettASwift: Going to leave this one for you...

source/view/blocks.js
```diff
@@ -1673,12 +1675,12 @@ Blockly.JavaScript['draw_triangle'] = function(block) {
 
   var moveC = constructBlockExeFuncCall( block, actionC, 'moveRadial' );
 
-  var start = "vwf.fireEvent( '" + vwf_view.kernel.application() + 
-                  "', 'blockExecuted', " + " [ 'startTriangle', '" + block.id + "', '" + block.data + "', " + 1 + " ] );\n";
-  var end = "vwf.fireEvent( '" + vwf_view.kernel.application() + 
-                  "', 'blockExecuted', " + " [ 'endTriangle', '" + block.id + "', '" + block.data + "', " + 1 + " ] );\n";
-  var mark = "vwf.fireEvent( '" + vwf_view.kernel.application() + 
-                  "', 'blockExecuted', " + " [ 'markPoint', '" + block.id + "', '" + block.data + "', " + 1 + " ] );\n";
+  var start = "vwf.callMethod( '" + vwf_view.kernel.application() +
```
BrettASwift: What context is this called in? We should not be accessing `vwf` directly. Would `this` be the blockly driver? Can we use `vwf_view.kernel` instead?

AmbientOSX: Yes, this is the Blockly driver. I have tried using vwf_view.kernel with mixed success in the past. This seems to guarantee that the method is called or event fired. I am not sure why.

BrettASwift: Are you able to use `this.kernel.callMethod` then? Have you tried that?

AmbientOSX: I haven't tried that yet, I'll give it a shot

AmbientOSX: Yeah... neither vwf.kernel, nor vwf_view.kernel, nor this.kernel work with the current configuration of the blockly driver. Im wondering if there is a quick change that could be made to set the right scope... maybe relating to this line ''' interpreter.setProperty( scope, 'vwf', myVwf ) ''' at line 507 of blockly.js model-side?

AmbientOSX:
Unless there are objections I am going to go ahead and merge this @BrettASwift - we can take a look at replacing the direct vwf call next week?

BrettASwift:
Yeah, that's fine.

PR #609, opened 2015-07-08
Nested Blockly Toolbox

jyucra:
Changed createTriangle and the forLoop block so that they have pre-built inputs set to 0 even while in the toolbox. This cuts down on the number of blocks you have to drag over.

### DIFF
--- a/source/scenario/blockly/scenario2f.xml
+++ b/source/scenario/blockly/scenario2f.xml
@@ -2,7 +2,7 @@
   <category name="Control">
   <block type="rover_moveForward"></block>
   <block type="rover_turn"><field name="DIR">Turn: Left</field></block>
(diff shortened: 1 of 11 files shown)


source/scenario/blockly/scenario2f.xml
```diff
@@ -2,7 +2,8 @@
   <category name="Control">
   <block type="rover_moveForward"></block>
   <block type="rover_turn"><field name="DIR">Turn: Left</field></block>
-  <block type="controls_repeat_extended"></block>
+  <block type="controls_repeat_extended" id="69" inline="true" x="60" y="100"><value name="TIMES"><block type="math_number_field" id="74" inline="false"><field name="VALUE">0</field></block></value></block>
```
kadst43: For this scenario only, let's set the value to 3.  In all other places, it should be 0.

kadst43:
Once you make the change to scenario 2f, :+1:

PR #611, opened 2015-07-08
Mission 4 blockly

jyucra:
Added pre-built blockly blocks to tasks 2 through 6 in order to make the player use extra operation blocks instead of setting the new vertex values by hand.


kadst43:
:+1:

PR #612, opened 2015-07-09
Grid Visual Tweaks

BrettASwift:
@kadst43 @AmbientOSX 

Made the grid lines thinner and less intense so the red and blue lines are more apparent. Also changed the normal grid lines to be closer to white where they were gold/yellow.


kadst43:
Definitely better.  Would be nice to make the red line even brighter - it's awfully close to the ground color - but if that's not easy, this is good for now.

:+1:

PR #613, opened 2015-07-09
Blockly Block Changes & Some VO

AmbientOSX:
@kadst43 Changes the divide block, removes sensor readings (though leaves variable readings for the moment, let me know if that makes sense to change), and changes some Perry VO subtitles to [whistles]


kadst43:
:+1:

PR #614, opened 2015-07-09
Blockly Refactor

AmbientOSX:
Requires https://github.com/virtual-world-framework/vwf/pull/505

@kadst43 @BrettASwift @jyucra 

All necessary changes are made and procedures can be defined in the toolbox and workspace now. Lingering issue with the toolbox div appearing that I've tried to fix with CSS multiple ways but Im thinking maybe you have a better idea @BrettASwift ?

Will continue working at the problem regardless.

### DIFF
--- a/source/view/blocklyUI.js
+++ b/source/view/blocklyUI.js
@@ -16,7 +16,6 @@ var ramBarCount = document.createElement( "div" );
 var ramBar = document.createElement( "div" );
 var currentRam = document.createElement( "div" );
 var startBlocklyButton = document.getElementById( "runButton" );
-var blocklyToolboxDiv = document.getElementById( "blocklyToolboxDiv" );
 var blocklySpeedButton = document.createElement( "div" );
 
 function setUpBlocklyPeripherals() {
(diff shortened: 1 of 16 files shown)


source/view/blocklyUI.js
```diff
@@ -31,6 +31,9 @@ function setUpBlocklyPeripherals() {
     var procedureIndicator = document.createElement( "div" );
     var indicatorHighlighter = document.createElement( "div" );
 
+    var toolboxDivs =  document.getElementsByClassName( "blocklyToolboxDiv" );
+    var blocklyToolboxDiv = toolboxDivs[0];
```
BrettASwift: This is undefined. It seems the blockly toolbox is not being created with the main blockly window.

BrettASwift:
The biggest issue seems to be that the timing is off. We need to find a way to wait until the toolbox is created before trying to append it to the wrapper.

BrettASwift:
Actually, that may be the second biggest issue. Even if we append it, it seems that blockly is set up to apply the position information directly to the element itself, which gets the highest priority and overrides our CSS rules. How we're doing it now may not work.

AmbientOSX:
I've rolled back to a stable version of Blockly and made a note of the commit. This are working and I am testing. Please be sure to update both this side and VWF side to test @jyucra https://github.com/google/blockly/commits/67140282a08b10d0beb8d842a2dc35beadc5d5e0

kadst43:
Can we get this safely merged in in time for the playthrough today?

AmbientOSX:
Kevin says :+1:

PR #617, opened 2015-07-13
Camera Pan Fix

BrettASwift:
@kadst43 @AmbientOSX 

I didn't think this would take so long to figure out, but this should do the trick. When panning to the node, we're copying the camera's current pose and world offset so no matter what mode/rotation is on the camera, it will be preserved.


kadst43:
:+1:, merging

kadst43:
I merged but didn't delete the branch.

PR #619, opened 2015-07-15
Editor

BrettASwift:
@kadst43 @AmbientOSX 

I'm going to be making changes to the editor in the new terrain branch so I can use the tool to generate the tile map. The tile map tool isn't complete, but it is functional. I'd like to get this merged. The old editor doesn't work anymore, so this isn't breaking anything.


AmbientOSX:
:+1:

PR #620, opened 2015-07-15
New mission 3 with procedures

jyucra:
Changes the old temporary mission 3 with a new mission 3 which introduces the player to procedure blocks. I also changed blockly so that it only allows the player to start the program if there is a single "topBlock" present in the workspace, where topBlock is any non-procedure-definition block that has no parent.


AmbientOSX:
:+1:

PR #623, opened 2015-07-16
Draw nanites

BrettASwift:
@kadst43 @AmbientOSX 

This adds the particle systems for drawing the triangles. Performance may be a concern since the particles are being handled on the CPU. Slower machines may get bogged down and we may have to reduce the particle count and rate of regeneration. However, give it a try and see what you think.

### DIFF
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -87,6 +87,7 @@ methods:
   addAxisOffset:
   removeAxisOffset:
   createNaniteSystem:
+  deleteNaniteSystem:
 events:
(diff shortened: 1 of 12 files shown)


AmbientOSX:
Looks good to me.

BrettASwift:
The whole particle system is generated at once and spawns particles randomly between two points. The stop point moves with the rover until the rover reaches the second point in the triangle. So, performance will not change with the rover speed. Shorter lines will be more densely populated with particles, however.

PR #625, opened 2015-07-21
Rover tracks and heightmap

BrettASwift:
@kadst43 @AmbientOSX 

- Added rover tracks
- Updated heightmap with ramp


AmbientOSX:
:+1:

PR #629, opened 2015-07-22
New sounds and other changes

jyucra:
* Added spray can noise during solar panel construction
* Added new sounds for the rovers
* Got rid of low ram limits in mission 4
* Switched the fill-in number blocks for drop-down menu number blocks.
* Fixed problem in mission 2 that switched you out of the graph tab between tasks


kadst43:
:+1:

PR #637, opened 2015-07-27
Changes to Mission 3/4 plus callOutObjective Fix

jyucra:
* Made changes to mission 4 so that it builds the solar panels more towards the middle
* Mission 3 ends in a new location; the base pod was moved south west
* Edited mission briefs/vo text/objectives so that they give the correct instructions
* Fixed a bug with callOutObjective that caused the blinking tile to be created in the wrong location/made it move around along with the camera.


kadst43:
:+1:

PR #638, opened 2015-07-29
New integer range and other fixes

jyucra:
* Expanded drop-down integer block range 
* Added blinking tiles to m3t3.
* Edited blockly categories
* Updated the sounds associated with solar panel construction
* Made some vo-text changes
* Switched the starting scenario back to the cinematic intro


kadst43:
If you haven't already, make sure you test that nothing broke... then :+1:

PR #639, opened 2015-07-29
Makes sure "mission critical" UI buttons always load

AmbientOSX:
@BrettASwift @kadst43 

This is currently in the study branch but I never PR'd for dev. Let me know if this is stylistically kosher or whether we should just set the draw to be 'onload' based from the start rather than using if/else

### DIFF
--- a/source/hud/blocklyButton.js
+++ b/source/hud/blocklyButton.js
@@ -15,6 +15,8 @@
 this.draw = function( context, position ) {
     if ( this.icon ) {
         context.drawImage( this.icon, position.x, position.y );
+    } else {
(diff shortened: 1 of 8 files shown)


kadst43:
@BrettASwift: I'll leave this one for you. :)

BrettASwift:
I'll have to test this. This would work if the issue was caused by the draw happening before the icon loaded. However, the draw call happens repeatedly, as fast as the browser allows. So, if it isn't drawn the first time, whenever it is loaded, that draw call will draw it. I'll do a little debugging and see if I can find what's causing the problem or if this seems to resolve the issue.

PR #640, opened 2015-07-29
Splits the and/or/not block into constituent blocks

AmbientOSX:
@kadst43 We wanted this a long time ago but I was checking my todos against what is in game and it never made it in somehow so here it is.


kadst43:
I'm not sure we ever actually use these... but this is fine to check in.

PR #641, opened 2015-07-29
Faulty position reading fix and more

jyucra:
* Fixed a bug with the position sensors which caused failures to happen in Mission 3 even though the right procedure was provided.
* Made changes to the procedure tab.
* Adds the new VO featuring Allison as the MC voice.


kadst43:
:+1:

PR #643, opened 2015-07-30
Changes before new players

jyucra:
* Adds limits to mission 4 (both ram and blocks wise) that forces the player to solve the scenario in the way we want them to.
* Moves blinking tile down the the rover's current position


kadst43:
:+1:

PR #644, opened 2015-07-30
Pick up method change and other things

jyucra:
* Changes the way that the supplies are picked up (done by setPickupActive now)
* Disabled Manny's tab in Mission 4 since he shouldn't do anything
* Got rid of the Cartesian Plane VO in Mission 2 since it no longer fits now that the VO prior to it has been fixed.


kadst43:
Could you comment out the unused VO in MarsGameSounds?  Then :+1:

PR #645, opened 2015-07-31
Fog

BrettASwift:
@kadst43 @AmbientOSX 

I added the fog to the scene. I was adding it to the VWF core when I saw someone already added it in recently, so I used what was there. I did have to make a change to allow the shader materials to get the fog define from threejs, so that is required for this. I also had to add code to our shaders to handle the fog.


AmbientOSX:
:+1:

PR #646, opened 2015-08-01
Blockly Core Changes

AmbientOSX:
@kadst43 @BrettASwift @jyucra 

Fixes some XML issues, splits conditional blocks, adds support for Blockly category icons. Still looking for a way to remove extra info from the right click menu (Help URL/Comments)... and hiding the right click menu/re-centering Blockly requires an SVG-redraw which should be occurring automatically but isn't so I'm working to determine when it would be best to make that call.

Depends on https://github.com/virtual-world-framework/vwf/pull/508


BrettASwift:
:+1:

PR #649, opened 2015-08-04
hud-draw-order

BrettASwift:
@kadst43 @AmbientOSX 

I finally got the bug to occur and promptly dug into the objects to find that the automatic draw order of the HUD was failing us. We have the battery element listed in `marsGameHud.vwf` first so that it will be the first object created. We do this because it has some draw commands that work as a mask. Drawing it after the other elements would cause them to disappear. It turns out that some times the battery element doesn't get created quick enough by VWF and the next elements in the list jump ahead, causing their draw order to be before the battery element's mask. So, to remedy this, I've made it possible to explicitly set the draw order of the HUD elements and added their draw orders to the `marsGameHud.vwf` file.

Requires: https://github.com/virtual-world-framework/vwf/pull/512

### DIFF
--- a/source/hud/blocklyButton.js
+++ b/source/hud/blocklyButton.js
@@ -15,8 +15,6 @@
 this.draw = function( context, position ) {
     if ( this.icon ) {
         context.drawImage( this.icon, position.x, position.y );
-    } else {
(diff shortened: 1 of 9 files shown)


AmbientOSX:
Very cool :+1:

PR #653, opened 2015-08-07
Perry Movement Speed

AmbientOSX:
@kadst43 @BrettASwift 

We disabled doing things this way because we thought it might be contributing to the wandering issue. That doesn't seem to be a problem anymore and the alternative way I was trying to fix/refactor this code was plagued by timing issues and cache issues. Lets do it this way for now since it is tried and true. I will revisit my refactor if I have time later next week.


kadst43:
Make sure you test a bit to see if it introduces significant wander.  If not, :+1:

PR #655, opened 2015-08-07
XML changes

jyucra:
* Sorted blocks out into the correct categories for the scenarios were this was still missing
* Made changes to the available blocks in Mission 3 so that you will only find a block in your toolbox if you will need it to complete that scenario.


AmbientOSX:
Set mission 3 task 7 procedure definition to be collapsed (in the xml set collapsed="true") then :+1:

jyucra:
I've rebased the branch to solve conflicts and made a couple extra changes to account for the separation of the conditional blocks. Another look would be appreciated.

source/scenario/newmission3task4.vwf.yaml
```diff
@@ -44,7 +44,7 @@ properties:
...
   - loadToolbox:
     - rover
-    - source/scenario/blockly/mission3b.xml
+    - source/scenario/blockly/mission3d.xml
```
kadst43: In ideal world, we'd name the files consistently as MissionXTaskY (no "new", no "mission3x")...

kadst43:
I added a small, nitpicky comment that doesn't need to be part of this PR... but in the meantime, let's get these changes in so that we can start playing with them.

PR #656, opened 2015-08-10
Sanity Check #1 Playthrough Changes

AmbientOSX:
Remove IDs from predefined XML (this breaks a lot of things), change Blockly Window Opacity, limit ability to move predefined XML, only open categories in early scenarios, move Blockly tooltip inside Blockly window.

@kadst43 @jyucra @BrettASwift


kadst43:
:+1:

PR #663, opened 2015-08-13
Double failure fix and scenario specific failures

jyucra:
* Fixes the double failure bug
* Adds a new trigger which checks whether a certain type of block is present or not in the workspace when blockly starts
* Adds scenario specific failures to Mission 2 that check whether they're using while/for loops, and fails them if they're not.
* Adds a scenario specific failure to m3t3 that checks whether an if block is present (I might make a more specific failure for this scenario, but it will require a different trigger to be made).


source/scenario/globalScenarioTriggers.vwf.yaml
```diff
@@ -138,6 +138,8 @@ properties:
           - readBlackboard:
             - stepThroughProgramHint
       actions:
+      - writeToBlackboard:
+        - disableFailOnIncompleteProgram
```
kadst43: I'm a bit concerned about this fix... among other things, it relies on the triggers firing in the correct order (which is why we implemented the trigger groups in the first place).  It also only solves this immediate problem.  As we add custom failures, those might have similar problems, and since this doesn't address the root cause of the problem it won't fix those issues.  And, of course, it runs the risk that the other failure won't get re-enabled (though I don't think that's the case).

The right way to fix this is to dig into the trigger group code, and figure out why the trigger group isn't preventing the lower priority failure from occurring.

source/scenario/newmission3task3.vwf.yaml
```diff
@@ -145,13 +145,26 @@ children:
...
+          priority: 3
+          triggerCondition:
+          - or:
+            - onBlockMissing:
+              - rover
+              - controls_if_nomut
+          actions:
+          - scenarioFailure:
+            - incomplete
+            - "You must use an if block."
```
kadst43: Does this cause a double failure?

kadst43: Also, on all of these we should think about the error message.  Is there a more generic error message we can give (or else maybe a more specific one)?  I don't want to tell them what the failure check is so that they can find another way to work around it, I want them to have to solve the problem properly.

Let's sit down together and talk through which missions these are...

source/triggers/clauses/clause_OnBlockMissing.js
```diff
@@ -0,0 +1,76 @@
...
+
+    return true;
+}
+
+this.onBlockMissing = function() {
+
+    var workspace = Blockly.getMainWorkspace();
+    var presentBlocks = workspace.getAllBlocks();
+    var present = false;
+    for (var i = 0; i < presentBlocks.length; i++) {
```
kadst43: It's a small thing, but the VWF folks are sticklers for their particular coding style - which, among other things, means putting spaces inside of parentheses.

source/scenario/scenario3e.vwf.yaml
```diff
@@ -129,6 +129,25 @@ children:
...
+            - onBlockMissing:
+              - rover
+              - controls_whileUntil
+            - onBlockMissing:
+              - rover
+              - controls_sensor_metal
+          actions:
+          - scenarioFailure:
+            - incomplete
+            - "You must next the loop block inside the conditional loop block."
```
kadst43: next => nest?

Although, as I said above, we should think over these messages regardless...

source/triggers/clauses/clause_OnBlockMissing.js
```diff
@@ -0,0 +1,76 @@
...
+        this.logger.warnx( "onGenerated", "blocklyStopped event not " +
+                            "found for '" + object.name + "'." );
+    }
+
+    if ( object.blocklyErrored ) {
+        object.blocklyErrored = this.events.add( function() { this.reset(); }, this );
+    } else {
+        this.logger.warnx( "onGenerated", "blocklyErrored event not " +
+                            "found for '" + object.name + "'." );
+    }
```
kadst43: Am I reading this right - can this only fire while the program is running?  If so... *should* it work that way?

source/scenario/scenario2f.vwf.yaml
```diff
@@ -111,6 +111,18 @@ children:
...
+          group: successOrFailure
+          priority: 3
+          triggerCondition:
+          - onBlockMissing:
+            - rover
+            - controls_repeat_ext
+          actions:
+          - scenarioFailure:
+            - incomplete
+            - "You must use a loop block."
```
kadst43: I think we call it a repeat block?

kadst43: Switch to the RAM fix.

source/scenario/newmission3task3.vwf.yaml
kadst43: "You need to put your conditional loops inside of an if block, so that they will only execute if you need to move south."

kadst43: Bump the priority down below "I didn't make it."

kadst43: Make the message be the same "I didn't make it" message.

source/scenario/scenario3a.vwf.yaml
```diff
@@ -128,6 +128,22 @@ children:
...
+              - rover
+              - controls_whileUntil
+            - onBlockMissing:
+              - rover
+              - controls_sensor_metal
+          actions:
+          - scenarioFailure:
+            - incomplete
+            - "You must use the conditional loop block and the metal detected block."
+
```
kadst43: Bump the priority down.

source/scenario/scenario3c.vwf.yaml
```diff
@@ -120,6 +120,25 @@ children:
...
+            - onBlockMissing:
+              - rover
+              - controls_whileUntil
+            - onBlockMissing:
+              - rover
+              - controls_sensor_metal
+          actions:
+          - scenarioFailure:
+            - incomplete
+            - "You must nest the loop block inside the conditional loop block."
```
kadst43: Let's not do this.

source/scenario/scenario3e.vwf.yaml
kadst43: Let's not do this one either.

kadst43:
@jyucra: bump! :)

PR #664, opened 2015-08-13
Asset updates

BrettASwift:
@kadst43 @AmbientOSX 

- Updated mini rover tracks
- Cargo pod now switches after success video
- Added updated arrow model, still working on how to best use that


index.vwf.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and 
 # limitations under the License.
 
---- 
+---
```
AmbientOSX: Is the + a convention or typo or...? :)

AmbientOSX:
I made a rather silly comment that I deleted so ignore that and :+1:

PR #665, opened 2015-08-13
Hide HUD During Instructional Dialog

BrettASwift:
@kadst43 @AmbientOSX 

This now hides all of the HUD elements apart from the comms portraits at the beginning of missions.


kadst43:
:+1:

PR #666, opened 2015-08-13
Make the conditional on the "this one is tricky" hint work correctly

kadst43:
@jyucra: I tested a bit, but keep an eye on this as you test as well, please.  It should play in m2t9 if you're facing east (or south).  If it doesn't play in m2t9 then it should play in m2t11 if you're facing north (or east), but it should only play once overall.


AmbientOSX:
:+1:

PR #667, opened 2015-08-13
Rover position

BrettASwift:
@kadst43 @AmbientOSX 

- Added rover position readout to HUD below the battery readout
- Restricted each mission to a single blockly context (blockly node), removing multiple tabs
- Fixed a bug with the mission speed button

### DIFF
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -77,8 +77,8 @@ methods:
   unpauseGame:
   resetHUDState:
   selectBlocklyNode:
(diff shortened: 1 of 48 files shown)


kadst43:
We should be removing the mission speed button (actually, I believe that we already did).  Aside from that, :+1:

BrettASwift:
Yeah, there were just references remaining. It should say "missing speed button" not "mission speed button."

PR #668, opened 2015-08-19
Perry tex

BrettASwift:
@kadst43 @AmbientOSX 

Finally have the finished Perry assets in and all working. This also has the tracks update, however the second tracks are not appearing. After lots of digging, I found the issue with them and will have to get Ken to re-export them. I will swap that before merging.


kadst43:
:+1:

PR #670, opened 2015-08-19
Reopen Blockly Category

AmbientOSX:
Reopens the blockly tab when the scenario reloads the toolbox XML and adds a safety check for a Blockly node being present. I realize it would be nice to keep it open but Blockly recreates the toolbox tree when it loads new toolbox XML. I will look into changing this in the core.

@kadst43


kadst43:
:+1:

PR #671, opened 2015-08-19
Suspend Renderer During Videos

BrettASwift:
@kadst43 @AmbientOSX 

Using David's change to suspend rendering, this stops rendering when each of the videos are playing. The increased performance also exacerbated the issue with the scene showing up before the intro video plays. I tried fixing it and only saw a very small improvement. However, the changes I made were good changes, so I left them in. I still am trying to find a way to solve that problem.

Let me know if the video performance is better now that rendering is suspended. Remember to update to the latest vwf/mars-game-development before testing.

### DIFF
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -18,6 +18,7 @@ implements:
   - "http://vwf.example.com/blockly/manager.vwf"
   - "http://vwf.example.com/threejs/fogExp2.vwf"
 properties:
+  enableRenderer: true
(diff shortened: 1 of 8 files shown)


kadst43:
This fixed the intro video, but when I jumped ahead to the mission 1 success video it was still broken.  Also, the in-game music didn't play.

BrettASwift:
I retested. All of it works on my end. The music plays and rendering seems to stop.

PR #672, opened 2015-08-19
Four Triangle Shader

BrettASwift:
@kadst43 @AmbientOSX 

This allows up to 4 triangles to be drawn by the shader. In order to pass them in, you'll just have to add 3 more points to the `material.triangle` property.

`material.triangle = [ pa, pb, pc, pd, pe, pf ]` Would be 2 triangles, where pa, pb, and pc are the points of the first and pd, pe, and pf are the points of the second. Each point is of the format `[ x, y ]`. Anything beyond the 12th point will not be rendered by the shader.

I can go ahead and make changes to the code to make adding/managing points easier if you'd like. If so, I will wait until that is done before merging this.

### DIFF
--- a/source/scenario/mission4task1.vwf.yaml
+++ b/source/scenario/mission4task1.vwf.yaml
@@ -248,6 +248,7 @@ children:
             - onScenarioSucceeded:
             - onScenarioFailed:
           actions:
+          - clearTriangle:
           - setProperty:
             - rover2
             - surveyArray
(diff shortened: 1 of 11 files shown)


BrettASwift:
This should be ready for scenario creation. All you have to do is add more `drawTriangle` actions for additional triangles to be drawn. The `clearTriangle` action clears all of the triangles. Each additional triangle requires a `drawTriangle` action, just to be clear. You cannot simply add more vertices to the existing one.

BrettASwift:
Going to wait for review on this one. Make sure whoever is using this for scenario creation knows how to use it.

AmbientOSX:
:+1: Lets merge this and I'll rebase/implement in Jonatan's branch when I'm debugging the other multi-triangle issues.

PR #673, opened 2015-08-20
Double Failure

AmbientOSX:
@kadst43 

I guess I also removed this future(0) because of the possibility of executing a second check unnecessarily and that was contributing in a low frame rate situation because the other check wouldn't execute before this one would regardless of groups? Could be dangerously wrong but let me know.

### DIFF
--- a/source/triggers/groups/triggerGroup.js
+++ b/source/triggers/groups/triggerGroup.js
@@ -60,8 +60,9 @@ this.checkTriggers$ = function() {
     //  impossible - especially with a low frame rate.  In order to address 
     //  this, do another check after a future(0).
     if ( haveTriggerToFire ) {
-        this.future( 0 ).checkTriggersCallback$();
-    } else if ( this.isChecking ) {
+        //this.future( 0 ).checkTriggersCallback$();
+        this.checkTriggersCallback$();
+    } else if ( this.isEvaluating ) {
         // schedule the next check
         this.future( this.checkFrequency$ ).checkTriggers$();
     }


source/triggers/groups/triggerGroup.js
```diff
@@ -60,8 +60,9 @@ this.checkTriggers$ = function() {
     //  impossible - especially with a low frame rate.  In order to address 
     //  this, do another check after a future(0).
     if ( haveTriggerToFire ) {
-        this.future( 0 ).checkTriggersCallback$();
```
kadst43: The future(0) is there for a reason - and potentially important to avoid odd race conditions.  It basically says "wait until everything else has had a chance to go, and then check again."  

For example, imagine that the order of events in a particular frame is:

- Blockly program marks itself as done
- The trigger group checks its triggers
- The collision system sends an event telling us the rover has collided with an obstacle

When the trigger group runs, the collision system has not yet registered the collision, so the scenario would fail because the program ended (or might even succeed) when it should fail because of the collision.

source/triggers/groups/triggerGroup.js
```diff
@@ -60,8 +60,9 @@ this.checkTriggers$ = function() {
...
-    } else if ( this.isChecking ) {
```
kadst43: It looks like isChecking and isEvaluating are two different things.  isChecking enables and disables the trigger group.  If you set it to false, the group will stop checking its triggers entirely, and if you set it to true then it will start again.  isEvaluating tells you whether right now, at this very instant, the trigger group is in the process of doing a check.  Every time a check is done, it first sets isEvaluating to true, then checks all the triggers, then sets isEvaluating back to false.  This prevents the triggers from firing independently of the trigger group (in theory).

I don't think I added isChecking - maybe @BrettASwift did?  In any case, it should have a comment on it - please add one. :)  Also, it seems like the if ( isChecking ) should wrap the entire function (or just force an early exit up at the top), not just prevent you from scheduling the *next* check.  The way it is now, there will always be one more check after you disable it, which doesn't seem desirable.

In any case... this doesn't seem to be a proper fix, though it could indicate that the problem is related to isChecking being enabled/disabled...

AmbientOSX:
Closing because of pair session earlier today - Kevin has real fix

PR #674, opened 2015-08-25
Scenario specific failures and blinking tile fix

jyucra:
* Added the recommended changes for the scenario specific failures
* Fixed the inputs for callOutObjective so it works with the recent changes


AmbientOSX:
:+1:

PR #676, opened 2015-08-25
Reset Blocks, Tooltips, Locked down Unit Tests, Other Fixes

AmbientOSX:
I'm still trying to resolve the RAM updating issue. I believe this is due to the block count not appropriately updating on the Driver side when the workspace is updated. Wanted to get input on this since I've struggled for a while on it now.

@kadst43 @BrettASwift 

Depends on: https://github.com/virtual-world-framework/vwf/pull/516


kadst43:
:+1: Let's get this merged, and then I'll test it with the rest.

PR #677, opened 2015-08-25
Rover facing arrow

BrettASwift:
@kadst43 @AmbientOSX 

Added the rover facing arrows. I placed them under the rovers, just above the ground because the arrow was completely blocking the view of perry. Let me know what you think.


kadst43:
:+1:

PR #679, opened 2015-08-26
Mission 4 PR Rebased

AmbientOSX:
@jyucra 

@kadst43 @BrettASwift 

We let the other PR get a bit out of sync so here is the new one rebased. We might want to review and get this in before it gets too much more unruly. I am looking at the M4T14 and nanite issues now but I could put those fixes in a new branch.


kadst43:
Make sure you've tested, then :+1:

PR #681, opened 2015-08-26
Mission 3 Changes

jyucra:
* Changes RAM and battery limits to more reasonable numbers
* Changed available blocks so that you only gain blocks and never lose them
* Added pauses between some VO
* Added comment blocks to test programs


source/view/blocks.js
```diff
@@ -3291,6 +3291,50 @@ Blockly.JavaScript[ 'graph_set_y' ] = function( block ) {
...
 };
 
+Blockly.Blocks['comment_block_test_program'] = {
+  init: function() {
+    this.appendDummyInput()
+        .appendField("This test program will confirm that your Navigate");
+    this.appendDummyInput()
+        .appendField("procedure works properly. Changes should be made");
+    this.appendDummyInput()
+        .appendField("to the procedure below, not to this test program.");
```
kadst43: Change "below" => "(below)"

kadst43:
One minor change, then :+1:

PR #682, opened 2015-08-26
Allow multiple blockly tabs

BrettASwift:
@kadst43 

Allows mutliple tabs to be enabled. I didn't do it for all scenarios, just one to test. Remember: the first node in the list will be set as the active one.

### DIFF
--- a/source/scenario/mission2task11.vwf.yaml
+++ b/source/scenario/mission2task11.vwf.yaml
@@ -69,6 +69,7 @@ properties:
     - [ -17, 32 ]
   - setBlocklyContext:
     - rover
+    - graph
   - loadToolbox:
     - rover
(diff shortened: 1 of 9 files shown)


kadst43:
@AmbientOSX: Could you (or Jonatan) use this to re-enable the graph tab in M2?

BrettASwift:
I will make the changes and update the PR.

BrettASwift:
I added the graph tab to all of the scenarios in mission 2 that you are moving the rover in so you can refer to the graph tab if needed. I did not add the rover tab to the scenarios where you need only the graph.

kadst43:
Sounds good.  We'll try it like this, and I see how to tweak it if I need to.

:+1:

PR #683, opened 2015-08-27
New assets and other changes

jyucra:
@kadst43 @BrettASwift 

* Adds new success videos for missions 2 and 3
* Adds new assets for the greenhouse, tent, and water well
* Gets rid of negative numbers and changes mission RAM/blocks to account for this
* Adds blinking tiles and triangles to Mission 4
* Adds scenario specific failures to mission 4 that check whether the nano spray block is present
* Changes the rover's display names to Perry/Peregrine/Rosie instead of rover/rover2/rover3 during low ram/out of ram messages.

There seems to be a problem with the drawTriangle action, as it only draws all 4/2 triangles if I switch directly into the scenario, but will only draw one if I play through the game in the correct sequence. There's also times where I'll call cancelCallOut during startState but the blinking tile will still be present.

@BrettASwift Can you confirm that I'm drawing multiple triangles the right/wrong way? I'm doing

  - clearTriangle:
  - drawTriangle:
    - [ -7, -13 ]
    - [ -7, -17 ]
    - [ -3, -17 ]
  - drawTriangle:
    - [ -9, -17 ]
    - [ -9, -13 ]
    - [ -13, -17 ]
  - drawTriangle:
    - [ -7, -19 ]
    - [ -7, -23 ]
    - [ -3, -19 ]
  - drawTriangle: 
    - [ -9, -19 ]
    - [ -9, -23 ]
    - [ -13, -19 ]


kadst43:
Let's get it merged, and address the bugs in a separate branch.

:+1:

PR #688, opened 2015-08-31
Brett's Punchlist

BrettASwift:
@kadst43 @AmbientOSX @jyucra 

- Made the mission brief window shorter for smaller resolution screens. The contents now have the ability to scroll.
- Fixed a console warning caused by the rover facing arrow changes.
- Added `setConstructed` action for immediately setting base components to constructed or hidden.
- Modified Mission 4 to use `setConstructed` action to fix objects that should be built not appearing.
- Fixed bug where blinking tiles would stay lit up after scenario changed.
- Fixed bug where highlighted tiles and triangles would stay visible when switching to a scenario without them.

### DIFF
--- a/source/scenario/mission4task1.vwf.yaml
+++ b/source/scenario/mission4task1.vwf.yaml
@@ -118,24 +118,20 @@ properties:
     - [ 21, -18, -10 ]
 
   #Solar panel construction
-  - setProperty:
+  - setConstructed:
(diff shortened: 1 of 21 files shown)


kadst43:
:+1:

kadst43:
Merging to get them in for my next test run.

Don't forget to delete the branch if you're done with it. :)

PR #689, opened 2015-09-01
- use () instead of [] for ordered pairs

kadst43:
- mission 3 gives only the degrees number block, not the drop-down one
- a few mission brief typos fixed
- started working through mission objectives & briefs for M4


BrettASwift:
:+1:

PR #693, opened 2015-09-02
Punchlist 9-2

AmbientOSX:
@kadst43 @BrettASwift 

Addresses the following punchlist items:

Spencer: In mission 3, the test programs should be locked down so that you can’t add blocks to them (either to the end, or on to the number blocks for the arguments, or anywhere else).
Spencer: The bug where Perry’s position wanders, and thus some triangles fail incorrectly, seems to be back.  Maybe related to the speed fix?  Coordinate with Jonatan for details…
Spencer: M3T2: I can say Direction of Travel != 270, but I can’t say 270 != Direction of Travel.  The != block won’t connect to the right-hand side of the degrees block.
o	Position X and Position Y blocks: “The [X/Y] value of the rover’s current position.”
o	y-coordinate block: “One of the parameters that was specified in the procedure call.”
o	x-coordinate block: same as y-coordinate, but for X.
o	Draw Triangle: “A block that sprays nanites in a triangle on the ground in order to construct a portion of your base.”
o	The tooltip issues that Mark sent (separate e-mail)
o	The degrees block tooltip doesn’t appear in the bottom right.
o	The conditional loop block tooltip sometimes appears on the block, sometimes in the bottom right, and sometimes both.
o	The if block tooltip doesn’t appear in the bottom right. (Should be working...)
Spencer: I can still move the test program
Spencer: M4 number block limitations:
o	M4T8: allow only 0, 1, and 7.
o	M4T9: allow only 0, 1, 3, and 7.
o	M4T10: allow only 0, 1, 2, 3, 7, and 9.
·Spencer (if quick): In M3, make both the procedure call and the procedure definition block unmovable.


kadst43:
I that all? ;)

:+1:

PR #696, opened 2015-09-03
Brett's Punchlist (9/3)

BrettASwift:
@kadst43 @AmbientOSX 

- Updated rover arrow model (Ken shortened it)
  - Moved it forward so it doesn't contact the rovers
- Removed Rosie (rover3) and references to her
- File cleanup
  - Removed old terrain assets that have been replaced
  - Removed original debris field models from phase 1
  - Removed old `gridObstacle` nodes
  - Removed unused parachute model
- Hide Blockly window when HUD enabled is set to false
- Enable the blinking tile even when tiles are not displayed
- Set building position, rotation, and scale in the scene instead of scenarios
  - Fixes the bug with the camera being pushed in at the beginning of the game
- Replaced Manny icons with teal versions

Kevin, here are a few issues you can add to my punchlist if you want:
- Fix solar panels 2, 3, and 4 being below the ground
- Add tile shader to crash pod ramp
- Move the battery that sits below the ramp to it's proper location
- (If quick) Investigate/fix shader console errors


AmbientOSX:
:+1:

PR #702, opened 2015-09-18
Fix y-coord bug when dragging from toolbox

AmbientOSX:
@kadst43 @BrettASwift


source/scenario/blockly/mission3c.xml
```diff
@@ -32,6 +32,6 @@
   </category>
   <sep></sep>
   <category name="Procedures">
-    <block type="variables_get" inline="false"><field name="VAR">y-coordinate</field></block>
+    <block type="variables_get" inline="false" editable="false"><field name="VAR">y-coordinate</field></block>
```
kadst43: Could we change the name of these while we're messing with them?  Destination X and Destination Y...

kadst43:
One comment (skip it for now if it's not a 30 second change), then :+1:

AmbientOSX:
Its a bit trickier because I need to test that the whitespace character works in predefined Blockly XML and then make changes to the scenario XML supporting it - I'll take a look in a separate PR.

PR #703, opened 2015-09-21
Brett punchlist 2

BrettASwift:
@kadst43 @AmbientOSX 

Finally have a working hack to fix the solar panels after the "quick and easy" fixes had no effect. I wish I had a real solution or at least a real understanding of what the problem actually is, but I'm sick of this bug so it will have to suffice.

Also only showing the facing arrow in top down view consistently, removed battery from under ramp, and resized jplayer to 800x600.

Wish this PR had more meat, but the solar panel bug really ate up all of my time.


kadst43:
:+1:

PR #704, opened 2015-09-23
Triangle Changes

AmbientOSX:
@kadst43 @BrettASwift 

Actually might want to get this reviewed and merged in since it is a bit old... If we decide we want to merge it in I'll remove the example blocks from the mission4task1 xml. It also takes care of the small task of making the eq block available where the neq blocks are.

There are 3 types of blocks... normal, auto and fixed. Normal blocks require the user to validate the operation they are performing on the block. Auto blocks propagate these fields automatically and the fields are not user-editable. The fixed block (which should be extended in scenario XML by setting the field values for OPX and OPY to whichever values you want it to be fixed to... ala:

`<block type="triangle_operations_dilate_auto_fixed" inline="false"><field name="OPX">0</field><field name="OPY">0</field></block>`

This also extends to other blocks (normal/auto) if you want certain blocks to default to a field in a particular scenario.

If a similar block is desired for translate that is a quick add.

Also makes use of new techniques for Blockly field input validation that I will carry over to subsequent PRs as well as advanced Block mutating techniques, setting images on blocks, setting the color of the Block dynamically, etc.

Let me know if there are any questions.

### DIFF
--- a/source/scenario/blockly/mission3b.xml
+++ b/source/scenario/blockly/mission3b.xml
@@ -16,6 +16,7 @@
   <sep></sep>
   <category name="Conditionals">
     <block type="logic_cond_neq_out"></block>
+    <block type="logic_cond_eq_out"></block>
   </category>
   <sep></sep>
   <category name="Math">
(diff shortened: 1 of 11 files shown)


kadst43:
:+1:

AmbientOSX:
@BrettASwift For rendering purposes, the vertices are stored in a dictionary that uses the block id as a key. currentBlocklyTriangles is exposed on the view. If you want to change this to send the vertices to the model you need only change line 149-153 in Blocks.js to a vwf.setProperty and send them wherever you like.

PR #706, opened 2015-09-28
Triangle Changes Addendum

AmbientOSX:
@kadst43 @BrettASwift 

-Adds in reflect and reflect_auto blocks
-Only propagates user-input values for non-auto blocks
-Fixes multiple-top-block counting in all cases


AmbientOSX:
Ah, forgot about changing the dilate variant to support a single integer value... one moment...

kadst43:
:+1:

PR #707, opened 2015-09-28
- Fixed the bug that keeps blockly programs from starting.

kadst43:
- Deleted some scenarios we're not using.


AmbientOSX:
:+1:

PR #708, opened 2015-09-29
Timer Updates

BrettASwift:
@kadst43 

- Start timer when user inputs are enabled
- Hide timer when HUD is hidden
- Move expand/collapse to the bottom of the timer window


kadst43:
:+1:

PR #710, opened 2015-10-01
Brett punchlist 3

BrettASwift:
@kadst43 @AmbientOSX 

- Updated Perry texture
- New Cargo Pod
- New Mini Rover Tracks
- Moved Perry cam closer
- Don't copy third person cam position when switching to a new target
- Fix shininess on new cargo pod
- Update rover position when grid axes change (they should be accurate 100% of the time, now)


kadst43:
:+1:

PR #711, opened 2015-10-02
Mission Brief Window

BrettASwift:
@kadst43 

Forgot to do this before starting more complex things, but here it is. I moved it to mirror the blockly window. I had to remove some things like the banner image so it's extremely plain, but I can spice it up when we've settled on a home for the window. It's probably too tall, but I wanted to leave it the same size as the blockly window for review.

### DIFF
--- a/index.vwf.html
+++ b/index.vwf.html
@@ -61,30 +61,27 @@ limitations under the License.
 </div>
 
 <!-- HTML -->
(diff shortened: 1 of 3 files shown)


kadst43:
:+1:, merging.  Remember to delete the branch if you're done with it... :)

PR #712, opened 2015-10-08
Fix Camera Jitter

BrettASwift:
@kadst43 @AmbientOSX @eric79 

So, it turns out that if you use the `transformTo` method without a duration, the threejs view driver skips interpolation of that transform. I opted to go with setting the property rather than adding a duration, since the duration can cause repeated transforms to get jammed up if it is too long and look jumpy if it's too short. This fixes the camera jitter when following the target node, however, if the app is running slowly, you'll notice some lag/fighting when controlling the camera while it is following the target node.


kadst43:
:+1: 

Next time you're at the school, it's worth taking a look at this and seeing how bad the lag/fighting is.

eric79:
:+1: Brett, you're like a surgeon ... a tiny one-line fix to make everything better.  :smile:

kadst43:
Cue Weird Al.

PR #716, opened 2015-10-09
Kevin polish 4

kadst43:


source/triggers/clauses/clause_IsAtPosition.js
```diff
@@ -48,6 +50,8 @@ this.onGenerated = function( params, generator, payload ) {
         return false;
     }
 
+    this.useAxisOffset = !!params[ 2 ];
```
AmbientOSX: What's the wisdom on the !! here? Asking purely for academic purposes :)

source/view/blocks.js
```diff
@@ -1662,377 +1662,377 @@ Blockly.JavaScript['triangle_operations_dilate_auto_fixed'] = function( block )
...
+        Blockly.FieldTextInput.numberValidator), "CX")
+        .appendField(",")
+        .appendField(new Blockly.FieldTextInput('0',
+        Blockly.FieldTextInput.numberValidator), "CY")
+        .appendField(")");
+    this.setInputsInline(false);
+    this.setPreviousStatement(true, 'INPUT');
+    this.setNextStatement(true);
+    this.setColour(195);
+    this.setTooltip('');
```
AmbientOSX: Be sure to check the other blocks for the correct tooltip formatting here when the tooltip is ultimately added.

AmbientOSX:
A few non-blocking questions then :+1:

source/triggers/clauses/clause_IsAtPosition.js
kadst43: It forces it to be a Boolean.  It's really not necessary in this case, but I felt like it added clarity (and prevents anybody from abusing the value, I suppose).

PR #718, opened 2015-10-11
Kevin polish 4

kadst43:


source/scenario/mission2task6.vwf.yaml
```diff
@@ -107,7 +107,6 @@ children:
             - "Use a loop to move along the slope of the tracks two more times."
           - setThirdPersonStartPose:
             - [ 16.316, 65, -15 ]
-          - startTimer:
```
AmbientOSX: If startTimer is outside of the generic onScenarioStart (and instead in the starting VO trigger) then the timer won't change if you ever progress then move back to a scenario (given the doOnce clause at line 114).. that's what was causing the timer issues I was experiencing Friday.

kadst43: Yeah... I'm pretty sure that that's actually what we want, though.  If it takes 5 tries to get through a scenario, the timer should count the total time.

With that said... we should keep an eye out for issues that come out of this.

AmbientOSX:
Just a quick note about the startTimer thing but maybe something to fix in another PR - :+1:

PR #719, opened 2015-10-12
Move the clearing of scenario state into the scenario's start function, instead of doing it in the scene.js

kadst43:
@BrettASwift: For sanity check.  I'll finish cleaning up before I merge it.


kadst43:
@BrettASwift: This is ready now "for real."

BrettASwift:
Did you test if this works as intended when failing a scenario? That's my only concern. If that works fine, then :+1:

PR #720, opened 2015-10-12
- The DoOnce trigger is now reset when you start a new game.

kadst43:
@BrettASwift: Could you sanity check and make sure that there's not a better way for me to do this?


BrettASwift:
If it's doing what I think it's doing, I guess it is fine.

PR #721, opened 2015-10-13
M4 Punchlist Items 10-13 AM

AmbientOSX:
@kadst43 @BrettASwift 

Set [] to ()
Vertically align coordinates
Resolve issue with error bubbles showing automatically (not able to have them automatically show yet - working on that fix)
Clear errors after block is deleted from workspace
MoveTo block now produces an error when not supplied component blocks
Limits bad block connectivity for transformation and flow blocks


BrettASwift:
:+1:

PR #723, opened 2015-10-13
Fix action + add view functions for camera setting

BrettASwift:
@kadst43 

I left the change to that action in there. Haven't determined if it is obsolete yet.

Edit: `readCameraPose()` will display the current pose in `[ radius, yaw, pitch ]` and `writeCameraPose( radius, yaw, pitch )` can be used to set the camera pose (you can also move it with the mouse, of course, but this way you can test cleaner numbers).


kadst43:
:+1:, merging

PR #724, opened 2015-10-13
Kevin polish 4

kadst43:
Various tweaks (see commit messages)


BrettASwift:
:+1:

source/view/blocks.js
```diff
@@ -2517,7 +2517,7 @@ Blockly.Blocks['start_triangle'] = {
   init: function() {
     this.setColour(90);
     this.appendDummyInput()
-        .appendField("startTriangle");
+        .appendField("Start Triangle");
```
AmbientOSX: Changing this will break old scene.js code we no longer use for drawing triangles but I guess it is fine to do

PR #725, opened 2015-10-13
Toggle the graph and/or tiles off at mission start:

kadst43:
- M1: do nothing
  - M2: toggle off tiles
  - M3: toggle off both
  - M4: toggle off tiles (only)


BrettASwift:
:+1:

PR #726, opened 2015-10-13
- Tweak mission starting camera angles.

kadst43:


kadst43:
Self-approving on this one.  It's trivial.

PR #728, opened 2015-10-14
10-14 Punchlist

AmbientOSX:
See commits.

@kadst43 @BrettASwift

### DIFF
--- a/source/view/blocks.js
+++ b/source/view/blocks.js
@@ -150,8 +150,8 @@ Blockly.Blocks['triangle_flow'] = {
         .appendField(" 1 , 0 ", "CURRENTC")
         .appendField(") ");
     this.setInputsInline(false);
-    this.setPreviousStatement(true,'Triangle');
-    this.setNextStatement(true,'Triangle');
+    this.setPreviousStatement(true,'Normal');
+    this.setNextStatement(true,'Normal');
     this.setColour(345);
     var thisBlock = this;
     this.setTooltip( function() {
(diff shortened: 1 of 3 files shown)


source/view/blocks.js
```diff
@@ -4867,8 +4852,8 @@ Blockly.Blocks['init_nano_construction'] = {
   init: function() {
     this.appendDummyInput()
         .appendField("Start Construction");
-    this.setPreviousStatement(true);
-    this.setNextStatement(true);
+    this.setPreviousStatement(true,'Normal');
+    this.setNextStatement(true,'Normal');
```
kadst43: I'm curious... what does this do?

kadst43:
Cosmetically, looks much better.

This doesn't seem to address the issue where M4T14 doesn't succeed when it should - I assume that you expected that.

I'm seeing a lot of issues (playing through M4) where the START button is initially greyed, even though I have a good program.  I disconnect and reconnect a block, and the button comes un-gray.  I saw that on very nearly every scenario.  I think it's specific to this branch... maybe something to do with the RESET changes?

source/view/blocks.js
AmbientOSX: It makes it so these blocks can only connect to blocks with connection statements set to nothing (this.setPreviousStatement(true)) or blocks that have the same string (can also be an array of types)... so a this.setNextStatement(true,'Normal') will only allow blocks with their setPreviousStatement set to 'Normal' or nothing.

kadst43:
:+1:

kadst43:
(you probably want to rebase before merge)

AmbientOSX:
@kadst43 rebased - take a quick look again and then I'll merge..

kadst43:
:+1:

PR #733, opened 2015-10-16
UI Tweaks

BrettASwift:
@kadst43 

- Re-enabled transparency on Blockly window (recolored to better fit with logger and brief window)
- Fixed initial Blockly tab selection (tab was stuck in unselected state prior to manually selecting a tab)
- Tweaked Mission Brief font size to match logger box

### DIFF
--- a/source/view/index.css
+++ b/source/view/index.css
@@ -114,13 +114,13 @@ a:active {
 }
 
 .blocklySvg {
-    background-color: rgba( 100, 100, 100, 1.0 );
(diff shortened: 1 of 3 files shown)


kadst43:
:+1: 

What was the issue with the blockly tab?  I don't think I caught that...

BrettASwift:
It was just a cosmetic thing. It wasn't showing the style as being selected until mission 2 when you manually clicked the graph tab and then it worked. It had been that way for a long time, but I didn't address it because I was the only one who noticed and it was such a minor issue.

PR #735, opened 2015-10-20
Brett punchlist

BrettASwift:
@kadst43 @AmbientOSX 

- Hide mission brief when going back to menu
- Disable blockly pointer events when blockly window is hidden
- Eliminate rapidly blinking tile by checking if there is a future call already scheduled
- Updated to new terrain and mixmap


kadst43:
:+1:

PR #736, opened 2015-10-21
Punchlist and Triangle Vis Support

AmbientOSX:
@BrettASwift @kadst43 

Commits should be somewhat explanatory but will give a rundown.
-Disabled tab for the browser window to avoid the reloading issue.
-y= is now movable but not deletable for M2
-The triangle_flow and transformation blocks will show the right A'B'C' "prime" count depending on which transformation it is.
-Warnings are now split up on multiple lines
-If block should now accept a wider variety of inputs
-Critically for triangle vis: if the block is in the flyout it will not send its ID or vertices over to the view for visualization.

Let me know if there are any questions.

### DIFF
--- a/source/view/index.js
+++ b/source/view/index.js
@@ -1566,6 +1566,6 @@ $( document ).keydown( function( objEvent ) {
     if ( objEvent.keyCode == 9 ) {  //tab pressed
         objEvent.preventDefault(); // stops its action
     }
-})
+} );
 
 //@ sourceURL=source/view/index.js
--- a/source/view/index.js
+++ b/source/view/index.js
@@ -1562,4 +1562,10 @@ function writeCameraPose( radius, yaw, pitch ) {
 
 window.addEventListener( "resize", checkPageZoom );
 
+$( document ).keydown( function( objEvent ) {
+    if ( objEvent.keyCode == 9 ) {  //tab pressed
+        objEvent.preventDefault(); // stops its action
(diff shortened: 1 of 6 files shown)


kadst43:
Add the ' marks to the bottom of the Draw Triangle block, then kick it back to me to retest.

AmbientOSX:
@kadst43 revised

source/view/index.js
```diff
@@ -1567,4 +1567,10 @@ function writeCameraPose( radius, yaw, pitch ) {
 
 window.addEventListener( "resize", checkPageZoom );
 
+$( document ).keydown( function( objEvent ) {
+    if ( objEvent.keyCode == 9 ) {  //tab pressed
+        objEvent.preventDefault(); // stops its action
+    }
+})
```
BrettASwift: Nitpick: Semicolon missing at the end and a space between the function bracket and parenthesis.

It's interesting that this seems to work. I tested the same thing without jQuery and it had no effect. The other possible difference is the context. I used `window` key events as opposed to `document`.

kadst43:
When I tested, the flow block wasn't executing.  I was testing in m4t12.

AmbientOSX:
@kadst43 Rebased and tested on all showstoppers, looks like this is good to go. Give this little branch that could another shot :)

kadst43:
:+1:

PR #740, opened 2015-10-22
Inherit pointer behavior in Blockly

BrettASwift:
@kadst43 

This should fix pointer events being enabled in blockly after running a program.


kadst43:
:+1:

PR #741, opened 2015-10-22
Finished updating programs and start/end positions for M3... 

kadst43:
except that I'd still like to get comments onto the procedure.

I still will need to update VO, mission briefs, and possibly
objectives... next week.


AmbientOSX:
:+1:

PR #742, opened 2015-10-29
Triangle shader revamp

BrettASwift:
@kadst43 @AmbientOSX 

Okay, so... This is working. I had to strip it down to the bare minimum because nothing beyond that is working. The different triangle colors don't work for some reason. I have no clue why, so the user triangle color is hard-coded in the shader. I will keep hammering on it. If we have to draw multiple triangles in the same scenario, I will have to revisit and test that a lot. Please pull this and play through mission 4 with it. If you run into any problems, let me know. I'm not super confident that this is bug free (maybe that's just frustration from the parts I haven't gotten working yet, I don't know). I played through it and nothing broke, for what it's worth.

### DIFF
--- a/index.vwf.yaml
+++ b/index.vwf.yaml
@@ -754,10 +754,6 @@ children:
     includes: source/scenario/mission3task4.vwf
   Mission3Task5:
     includes: source/scenario/mission3task5.vwf
-  Mission3Task6:
-    includes: source/scenario/mission3task6.vwf
-  Mission3Task7:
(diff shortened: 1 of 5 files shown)


kadst43:
I’m good with hard-coded colors for now.  At this point, I think it's fair to say that we won't bring multiple triangles back.

I'll aim to take a look tomorrow.

-K

kadst43:
Also looks good.  :+1:

PR #743, opened 2015-10-30
Updated shader

BrettASwift:
@kadst43 @AmbientOSX 

Rewrote the grid shader. The axis color should be more obvious now.


kadst43:
Much better.  :+1:

PR #746, opened 2015-11-02
Replaced blockly ui buttons

BrettASwift:
@kadst43 @AmbientOSX 

Remade the blockly interface buttons. I tried a few different designs, some with icons, stroked text, etc. I settled on a much cleaner look with lightly colored text that I think looks pretty nice on the darker blockly window we have now.


kadst43:
Is the text really off-center vertically, or does it just look that way in the diff?

If the former, should it be?

:+1:

BrettASwift:
There's actually three white dots below the text. Forgot to mention that.

PR #748, opened 2015-11-03
Triangle flow changes

AmbientOSX:
@kadst43 @BrettASwift 

Commits tell the story - changes how blocks flow values in the triangle missions if they aren't filled in.

### DIFF
--- a/source/view/blocks.js
+++ b/source/view/blocks.js
@@ -323,10 +323,8 @@ var checkAndSetField = function ( block, field ) {
     do {
       if ( currentBlock.type === 'triangle_flow' ) {
         return unitTriangle[ field ];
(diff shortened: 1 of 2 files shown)


source/view/blocks.js
```diff
@@ -300,6 +300,38 @@ Blockly.JavaScript['triangle_flow'] = function( block ) {
   return overallCode;
 };
 
+var validateFieldValue = function ( value ) {
+  if ( isNaN( value ) || value === "" ){
+      return '_';
```
kadst43: Could we use ' ' instead of '_'?  It would probably look better...

kadst43:
:+1:

PR #749, opened 2015-11-04
Triangle flow changes fix

AmbientOSX:
@kadst43


kadst43:
Make sure that you test, then :+1:. 

Not urgent, but sometime when you have a minute you might try using a space rather than an underscore or empty string.  That will look like an empty string, but won't actually be one.

Also, what happens if you empty the field out and then hit enter?  Will it go back to _?  Is there a way to go back to _?

AmbientOSX:
It will go back to _

PR #751, opened 2015-11-04
Fix yaml for m2t4

AmbientOSX:
@kadst43


kadst43:
I think both of these are in mainline already?

AmbientOSX:
Odd...

kadst43:
I made the fixes in parallel. :)

PR #753, opened 2015-11-09
Polish items 11-9

AmbientOSX:
@kadst43 @BrettASwift 

Getting these changes in to address the showstopper and a few quick other things. M3T2 resets XML appropriately for me but need verification. Blockly window closes on execute...

Things I have investigated but I'm inclined to say "won't fix" to: setting the underscores to spaces in triangle transformation blocks resurfaces the undefined bug. Tooltips and input block validation for procedures was done BUT required reenabling the "special" procedures and variables fields in the Blockly toolbox. The "correct" procedure_callnoreturn block would also not reliably load. This has to do a complex Core Blockly process for activating procedures.


source/scenario/mission3task1.vwf.yaml
```diff
@@ -36,8 +36,6 @@ properties:
     - [ "assets/video/Success3_800x600.mp4" ]
 
   #blockly
-  - setBlocklyContext:
-    - rover
```
kadst43: Why are we removing this?  It's probably not necessary unless we jump back from m4, so not critical - but is it hurting anything?

source/scenario/mission3task2.vwf.yaml
```diff
@@ -37,12 +37,10 @@ properties:
...
... inline="false"><field name="VAR">Desired Y</field><value name="INPUT"><block type="logic_cond_out" id="m" inline="false"><field name="VALUE">&lt;</field><value name="INPUT"><block type="controls_sensor_position_y" id="n" inline="false"></block></value></block></value></block></value><statement name="DO"><block type="rover_moveForward" id="o"></block></statement></block></next></block></statement></block></statement></block><block type="comment_block_free_text_no_in" id="p" deletable="false" movable="false" editable="false" inline="false" x="0" y="-340"><field name="COMMENT">This program tests the Navigate procedure.</field><next><block type="procedures_callnoreturn_no_out" id="u" deletable="false" movable="false" editable="false" inline="false"><mutation name="Navigate"><arg name="Desired X"></arg><arg name="Desired Y"></arg></mutation><value name="ARG0"><block type="math_number_field_no_out" inline="false" movable="false" deletable="false" id="a63"><field name="VALUE">0</field></block></value><value name="ARG1"><block type="math_number_field_no_out" inline="false" movable="false" deletable="false" id="a60"><field name="VALUE">0</field></block></value></block></next></block></xml>
```
kadst43: I just made extensive changes to the m3 programs.  Not sure what your change is, but assuming that it's small we should have you merge it into that version.

kadst43:
I'm not sure what all of these changes are doing, aside from the position sensor drop-down (which we do want).

I don't think we want to close the blockly window when the program starts.  There is a green task to close the *palette*, but at this point I wouldn't even do that - if it turns out to be the wrong thing, it's getting to be too late to change it back.

I'm good with dropping the m4 change from "_" to " ".

I didn't understand which the other issues that you wanted to drop are.  FWIW, I enabled the drop-down in the variable blocks, which means that those have some of the other procedure functionality that I'd prefer to conceal... I think the drop-down is more important than concealing the functionality in that case.

Please make sure you're working blue/black tasks first.  If you have tasks on your list that aren't on the punchlist, let me know and I'll figure out where to add it...

source/scenario/mission3task2.vwf.yaml
AmbientOSX: The startXML was malformed. This was preventing m3t2 from loading its XML on the start. All I did was copy the blockly_xml to startXML to make it work. As long as the new changes address the issue we should be good.

PR #755, opened 2015-11-10
Block changes 11-10

AmbientOSX:
@kadst43 @BrettASwift 

As usual, commits tell the story. Transformation blocks tested in each scenario they are used in.


kadst43:
As part of this PR or another, could you update all of the m3 programs to use the dropdown position sensor?

:+1:

PR #759, opened 2015-11-13
Bretts list

BrettASwift:
@kadst43 

I didn't get the ground fit fixed and tested or the fix for the rover heading yet. This should eliminate any bugs caused by multiple clicks of the start/stop button and timing issues with that (there were a few perfect moments when you could run a program at the end of a scenario and succeed/fail). I also moved the dust, which is perhaps not necessary, but it's in there. I did at least remove one world transform retrieval that we weren't using, so it's doing it for 4 instead of 5.

### DIFF
--- a/source/fourWheeledRover.js
+++ b/source/fourWheeledRover.js
@@ -5,7 +5,7 @@ this.placeOnTerrain = function( pos ) {
         pos[ 1 ] - this.transform[ 13 ], 
         pos[ 2 ] - this.transform[ 14 ] 
     ];
-    var terrainPosCentroid = this.getTerrainPosUnderNode( this, deltaPos );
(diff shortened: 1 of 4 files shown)


kadst43:
Do the dust changes significantly improve either appearance or performance?
Did you thoroughly test (especially the button click changes)?

If the answer to both of these is "yes," then :+1:.

kadst43:
I'm good with waiting until Mission 3 to add the dust, but the new dust looks weird - it's sort of glowy, and clumpy.  Going to hold off until next week on the PR.

BrettASwift:
Removed the dust change.

kadst43:
Good to merge into development. Could you also merge it into rc-2-staging?

PR #763, opened 2015-11-19
Updated the success conditions on certain m2 tasks so that you can't easily…

kadst43:
succeed by graphing the wrong line (in particular, y=4 used to solve the
second segment).

@AmbientOSX: Can you sanity check?  This is going into the rc2 branch...


AmbientOSX:
:+1:

PR #764, opened 2015-11-23
Escape Opens Options Menu

BrettASwift:
@kadst43 @AmbientOSX 

Also made it so that the options menu closes when returning to the menu and cannot be opened from the menu.


AmbientOSX:
:+1:

PR #765, opened 2015-11-23
Fix for blinking tab in mission 2 task 2

BrettASwift:
@kadst43 @AmbientOSX 

I didn't dig into this one. On a hunch, I assumed the way we're defining what tabs we use in the scenarios now is clearing the tabs and the blinking after we tell it to make the tab blink, so I moved the blink call to after the blockly window is opened. I tested it and it worked, but I cannot verify that my hunch is correct, only that this solution worked. Hope that's good enough for now. If not, I will dig into it further.

### DIFF
--- a/source/scenario/mission2task2.vwf.yaml
+++ b/source/scenario/mission2task2.vwf.yaml
@@ -27,8 +27,6 @@ properties:
     - environmentWind
   - blinkHUDElement:
     - blocklyButton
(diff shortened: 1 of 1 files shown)


AmbientOSX:
Blinks for me, looks good

kadst43:
@BrettASwift: Could you merge this fix into RC2?

PR #767, opened 2015-12-07
Remove rover alerts

BrettASwift:
@kadst43 @AmbientOSX 

Removed the alerts for low memory/power and blocked. None of them are very visible and the ones that are, just confuse the player. Up to you if you want to merge it. It's minor.


kadst43:
:+1:

PR #769, opened 2015-12-11
No alerts + helicam radius

BrettASwift:
@kadst43 @AmbientOSX 

Remove alerts and extend helicam radius so that you can pan to see the radio at the default zoom level.


AmbientOSX:
:+1:

PR #772, opened 2016-01-19
Optimize load

BrettASwift:
@kadst43 @AmbientOSX 

I replaced the obvious finds that were looking for nodes with known locations. There are still a lot of finds happening at startup it seems, but I wasn't able to eliminate those without a significant refactor. On my desktop, the load time went from 44 seconds total to 34 seconds total, so about a ~10 second difference.


kadst43:
:+1:

PR #774, opened 2016-02-01
Replace Master with MGD

AmbientOSX:
@kadst43 @BrettASwift


kadst43:
Do a quick test after the merge, just to sanity check that nothing broke.

:+1:

PR #776, opened 2016-02-01
Cycle loading text and pulse text opacity

BrettASwift:
@kadst43 

This steps through the loading text sequence you sent me and stops on the last one. The step interval is 15 seconds. The text opacity also cycles between 50-100%. It's a little twitchy while the assets are loading, but it smooths out when the scenarios are being generated. I felt it was a little too static without the opacity cycling.


kadst43:
Can you switch to a version of the image that doesn't have the built-in text at the bottom?  Try one of these two:

http://redmine.virtualworldframework.com/attachments/download/887/mars_load5.png 
http://redmine.virtualworldframework.com/attachments/download/888/mars_load6.png

Also, maybe make the text bigger/bolder and see how it feels.

kadst43:
You might check with Ken and see if it's possible to match the exact RGB value of the red he used...

BrettASwift:
Updated the image, matched the text color to the lines. I made the slightly bigger, but the bold didn't look good because it conflicted with the thin red lines bordering the image.

PR #777, opened 2016-02-02
Remove scene flash

BrettASwift:
@kadst43 @AmbientOSX 

This should remove the scene flashing at the start of the game. It also causes us to immediately cut to videos, though. So the 400ms fade in doesn't happen. I don't think it looks bad. Definitely the benefit outweighs the cost.


BrettASwift:
Merging.

